PR #12, opened 2021-10-22
feat: post `INIT` and `GAME_OVER` events, switch from `process.env` to `globalThis`

mattbretl:


src/index.ts
```diff
@@ -42,30 +58,13 @@ export const Rune: RuneExport = {
...
-      }, 3000)
+    // When running inside Rune, runePostMessage will always be defined.
+    if (globalThis.postRuneEvent) {
+      globalThis.postRuneEvent({ type: "INIT", version: "1.1.1" })
+    } else {
+      mockEvents()
     }
   },
-  // Allow Rune to see which SDK version the game is using
-  version: "1.1.1",
```
bfelbo, line 68: It might be nice to keep the version as a separate exposed value that can easily be extracted w/o injecting `postRuneEvent` and looking at the `INIT` event. For instance, that might be useful as part of quality control and deployment scripts when game devs submit their games to Rune.

Similarly, it might be useful if we're debugging games that never run `init()` for some reason.

WDYT?

src/index.ts
```diff
@@ -76,7 +75,34 @@ export const Rune: RuneExport = {
   _pauseGame: () => {
     throw new Error("Rune._pauseGame() called before Rune.init()")
   },
-  gameOver: () => {
-    throw new Error("Rune.gameOver() called before Rune.init()")
+  gameOver: ({ score }) => {
+    if (!Rune._doneInit) {
+      throw new Error("Rune.gameOver() called before Rune.init()")
+    }
+    globalThis.postRuneEvent?.({ type: "GAME_OVER", score })
```
bfelbo, line 87: Since we're providing typing for other Rune codebases that `score` is a number, let's also validate that `score` is in fact a number. This might also be helpful for game devs when integrating the SDK and accidentally passing the wrong type.

bfelbo (approved):
Looks great! 💯

src/index.ts
mattbretl: 👍  I added it to `INIT` for our primary in-app use case, since we can't just reach inside and inspect `Rune.version` without injecting JS that posts a message to the app and adding a message handler on the app side, and at that point, we might as well just include it in `INIT`. But it's trivial to keep `Rune.version` for the other purposes you mentioned so I'll add it back in.

PR #15, opened 2021-10-28
build: only publish dist/

mattbretl:
Trivial change to only publish dist/ (and LICENSE and README.md and package.json, which are always included).


bfelbo (approved):
👍

PR #19, opened 2021-11-14
refactor: rename `getScore` functions

mattbretl:
It's a little confusing how `init(…)` accepts `getScore`, that `getScore` gets assigned to `_getScoreFromGame`, and then the Rune app calls an entirely different `_getScore`.

Let's refactor this internally to assign `getScore` to `_getScore` and call `_requestScore` from the Rune app.

`npm run build` and `npm run test` both run successfully.


bfelbo (approved):
💯

PR #20, opened 2022-02-03
feat: do not allow for external dependencies

mieszko4:
* Add title and favicon.
* Add pixi.min.js so that is not requested externally.
* Allow to use SDK only by including it in `index.html`
* Adds`Submission` section with information about:
  * zipping
  * having `index.html` as the entry point
  * having no external dependencies

Closes RUNE-6389


bfelbo (approved):
LGTM!

PR #26, opened 2022-02-11
feat: pass challenge number to game over event

mieszko4:
Pass challenge number to SCORE and GAME_SCORE events to get information for which challenge number the score applies to.


bfelbo (approved):
💯 🥳

PR #27, opened 2022-02-11
docs: challenge number

bfelbo:
Made a few suggested changes to try to:
- clarify that the challenge number can also be used for other gameplay content than maps
- make it more clear that challenge number is optional
- hide away the ways challenge number could be used in collapsible sections

It seems easier to review this by looking at the new markdown file than seeing line-by-line changes.


mieszko4 (approved):
🚀 Awesome!

It would be great to update the example's comments, too in https://github.com/rune/rune-games-sdk/blob/staging/examples/bunny-twirl/index.js#L63-L73 to mention `1) As a seed for randomness` and `2) Iterate through a fixed list of game content (e.g. maps)`

mattbretl (approved):
Looks great!

PR #29, opened 2022-02-23
refactor: internal types at bottom

bfelbo:
Minor refactoring that moves the internal types to the bottom of the file. This makes it easier for any new game dev to read the file as they'll only see relevant info for them in the beginning (`InitInput` and `RuneExport`).

Also added comments to explain why we need `declare global` and what these events are for.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -135,13 +135,13 @@ const RuneLib = {
   },
 }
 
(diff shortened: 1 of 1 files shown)


mattbretl (approved):

mieszko4 (approved):
👍

PR #32, opened 2022-03-02
build: fix typescript eslint

bfelbo:
We were missing `@typescript-eslint/eslint-plugin` so eslint wasn't working.

As part of fixing that, I noticed that we're intentionally breaking some rules as part of the SDK. We could add 11 `// eslint-disable-next-line` to ignore the errors in the code (and 5 more if we want to also ignore the console.log warnings). However, that's almost 1 per 10 lines of code. To keep the SDK code itself simple I instead disabled these rules.

It is nice to keep the SDK code simple, but these rules are also quite important ones that could easily catch mistakes in the future. I could go either way on this. WDYT @mattbretl @mieszko4?

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -12,8 +12,7 @@ module.exports = {
   rules: {
     "no-throw-literal": "error",
 
-    // Used to inject into global space
-    "no-undef": "off",
+    // Used for communicating with the game and Rune
     "no-unused-vars": "off",
 
     // Used to improve developer experience
@@ -28,5 +27,18 @@ module.exports = {
     Promise: "readonly",
     require: "readonly",
     setTimeout: "readonly",
+    globalThis: "readonly",
   },
+
+  overrides: [{
+    files: ["test/*.ts"],
+    globals: {
+      jest: "readonly",
+      describe: "readonly",
+      expect: "readonly",
+      test: "readonly",
+      beforeEach: "readonly",
+      afterEach: "readonly"
+    }
+  }]
 }
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,4 +1,5 @@
 module.exports = {
+  root: true,
   env: {
     es6: true,
   },
@@ -9,8 +10,14 @@ module.exports = {
   extends: ["eslint:recommended", "prettier"],
   plugins: ["@typescript-eslint"],
   rules: {
-    "no-console": ["warn"],
     "no-throw-literal": "error",
+
+    // Used to inject into global space
+    "no-undef": "off",
+    "no-unused-vars": "off",
+
+    // Used to improve developer experience
+    "no-console": "off"
   },
   settings: {},
   globals: {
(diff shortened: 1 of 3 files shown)


.eslintrc.js
```diff
@@ -1,4 +1,5 @@
 module.exports = {
+  root: true,
```
bfelbo, line 2: Recommended in https://typescript-eslint.io/docs/linting/#configuration.

package.json
```diff
@@ -25,10 +25,10 @@
     "url": "https://github.com/rune/rune-games-sdk/issues"
   },
   "homepage": "https://github.com/rune/rune-games-sdk#readme",
-  "dependencies": {},
```
bfelbo, line 28: Was deleted automatically.

mieszko4:
> It is nice to keep the SDK code simple, but these rules are also quite important ones that could easily catch mistakes in the future. I could go either way on this. WDYT @mattbretl @mieszko4?

1. `no-undef`: In `.eslintrc.js` you could define `globalThis` in `globals`. For our test folder you would also need to define globals like `expect`, `describe`, etc. In order to scope that only for test folder, you could use `overrides` key.
2. `no-unused-vars`: The problem seems to be typescript-specific. Adding "plugin:@typescript-eslint/recommended" in `extends` resolves this but brings bunch of new eslint errors :) So maybe it makes sense to keep it off for now.

mattbretl (approved):
I'm okay with changing these to "warn" or "off", whatever you prefer. 👍

mattbretl:
> 1. `no-undef`: In `.eslintrc.js` you could define `globalThis` in `globals`. For our test folder you would also need to define globals like `expect`, `describe`, etc. In order to scope that only for test folder, you could use `overrides` key.

Good call @mieszko4! That's the most important one, and sounds simple enough to add?

mieszko4:
> > 1. `no-undef`: In `.eslintrc.js` you could define `globalThis` in `globals`. For our test folder you would also need to define globals like `expect`, `describe`, etc. In order to scope that only for test folder, you could use `overrides` key.
> 
> Good call @mieszko4! That's the most important one, and sounds simple enough to add?

Yes, I played a bit with it 😄 You should be able to do:
```diff
diff --git a/.eslintrc.js b/.eslintrc.js
index 8c97438..08679d6 100644
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -13,7 +13,6 @@ module.exports = {
     "no-throw-literal": "error",
 
     // Used to inject into global space
-    "no-undef": "off",
     "no-unused-vars": "off",
 
     // Used to improve developer experience
@@ -28,5 +27,17 @@ module.exports = {
     Promise: "readonly",
     require: "readonly",
     setTimeout: "readonly",
+    globalThis: "readonly",
   },
+
+  overrides: [{
+    files: ["test/*.ts"],
+    globals: {
+      jest: "readonly",
+      describe: "readonly",
+      expect: "readonly",
+      test: "readonly",
+      beforeEach: "readonly"
+    }
+  }]
 }
```

bfelbo:
Amazing @mieszko4! Was hoping there would be a cleaner solution like this. Thanks!

PR #33, opened 2022-03-02
feat: add deterministicRandom()

bfelbo:
Needed to make it easier for game devs than using `seedrandom`. WIP.

Builds upon https://github.com/rune/rune-games-sdk/pull/32.

PR #38, opened 2022-03-03
feat: add deterministicRandom() using PRNG

bfelbo:
Cleaner implementation of https://github.com/rune/rune-games-sdk/pull/33. Exposes `deterministicRandom()` for game devs to use to generate maps etc. in a deterministic way.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -76,10 +76,7 @@ export const Rune: RuneExport = {
     }
     const score = Rune._getScore()
     Rune._validateScore(score)
-
-    // Reset randomness to be deterministic across plays
     Rune._resetDeterministicRandom()
-
     globalThis.postRuneEvent?.({
       type: "GAME_OVER",
       score,
@@ -89,7 +86,7 @@ export const Rune: RuneExport = {
   getChallengeNumber: () => globalThis._runeChallengeNumber ?? 1,
   deterministicRandom: () => {
     // The first time that this method is called, replace it with our
-    // determinstic random number generator and return the first number.
+    // deterministic random number generator and return the first number.
     Rune._resetDeterministicRandom()
     return Rune.deterministicRandom()
   },
@@ -145,10 +142,7 @@ export const Rune: RuneExport = {
     Rune.gameOver = function () {
       const score = Rune._getScore()
       Rune._validateScore(score)
-
-      // Reset randomness to be deterministic across plays
       Rune._resetDeterministicRandom()
-
       globalThis.postRuneEvent?.({
         type: "GAME_OVER",
         score,
@@ -192,6 +186,7 @@ export const Rune: RuneExport = {
     return seed()
   },
   _resetDeterministicRandom: () => {
+    // Reset randomness to be deterministic across plays
     Rune.deterministicRandom = Rune._randomNumberGenerator(Rune.getChallengeNumber())
   },
 }
--- a/src/index.ts
+++ b/src/index.ts
@@ -76,7 +76,12 @@ export const Rune: RuneExport = {
     }
     const score = Rune._getScore()
     Rune._validateScore(score)
-    globalThis.postRuneEvent?.({ type: "GAME_OVER", score, challengeNumber: Rune.getChallengeNumber() })
+
+    // Reset randomness to be deterministic across plays
+    const challengeNumber = Rune.getChallengeNumber()
+    Rune.deterministicRandom = Rune._randomNumberGenerator(challengeNumber)
+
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -123,16 +134,56 @@ export const Rune: RuneExport = {
     Rune.gameOver = function () {
       const score = Rune._getScore()
       Rune._validateScore(score)
-      globalThis.postRuneEvent?.({ type: "GAME_OVER", score, challengeNumber: Rune.getChallengeNumber() })
+
+      // Reset randomness to be deterministic across plays
+      const challengeNumber = Rune.getChallengeNumber()
+      Rune.deterministicRandom = Rune._randomNumberGenerator(challengeNumber)
+
+      globalThis.postRuneEvent?.({ type: "GAME_OVER", score, challengeNumber })
```
bfelbo: A bit messy to have to duplicate this with actual `gameOver()` logic, but couldn't think of a cleaner way w/o adding complexity.

src/index.ts
```diff
@@ -72,9 +76,16 @@ export const Rune: RuneExport = {
...
+
+    // Reset randomness to be deterministic across plays
+    const challengeNumber = Rune.getChallengeNumber()
+    Rune.deterministicRandom = Rune._randomNumberGenerator(challengeNumber)
+
+    globalThis.postRuneEvent?.({ type: "GAME_OVER", score, challengeNumber })
   },
   getChallengeNumber: () => globalThis._runeChallengeNumber ?? 1,
+  // @ts-ignore This will be set by onLoad()
+  deterministicRandom: undefined,
```
bfelbo: There's a test that covers that this gets set correctly.

test/sdk.test.ts
```diff
@@ -226,4 +226,42 @@ describe("sdk", function () {
     // See that challenge number is correct
     expect(challengeNumber).toEqual(123)
   })
+
+
+  test("deterministicRandom() works before init()", async function () {
+    const randomArray = [...Array(7)].map(() => Math.round(Rune.deterministicRandom() * 10))
+    expect(randomArray).toEqual([1, 4, 4, 3, 5, 1, 7])
```
bfelbo, line 234: Doesn't use snapshot as it gets quite long and I don't expect this to change.

mattbretl:
I created a branch that separates resetting the RNG (to avoid duplicating logic), removes `_onLoad`, and eliminates the need for `@ts-ignore`: https://github.com/rune/rune-games-sdk/tree/feat/reset-deterministic-random Replacing the method on first call doesn't feel especially clean 😏 , but it might be an net improvement overall?

EDIT: You'll also see some `prettier` changes. Those should probably be applied on this branch regardless.

bfelbo:
Haha, I have a half-baked branch going down the same path, but stopped at same consideration as you 😅 Your implementation looks great though, thanks! Sounds good to me to move ahead with that.

mattbretl (approved):

src/index.ts
bfelbo: Fixed by @mattbretl with https://github.com/rune/rune-games-sdk/pull/38/commits/46ad431e06aa54e01bda9529f5760279942f8862.

src/index.ts
bfelbo: Fixed by @mattbretl with https://github.com/rune/rune-games-sdk/pull/38/commits/46ad431e06aa54e01bda9529f5760279942f8862.

mieszko4 (commented):
🚀

test/sdk.test.ts
```diff
@@ -213,17 +213,64 @@ describe("sdk", function () {
...
+    const randomArray2 = [...Array(7)].map(() =>
+      Math.round(Rune.deterministicRandom() * 10)
+    )
+    expect(randomArray1).toEqual(randomArray2)
+  })
+
+  test("deterministicRandom() seed changes dramatically with every challenge number", async function () {
+    // This is important to prevent correlation etc. between challenges
+    expect(Rune._hashFromString("1")).toMatchInlineSnapshot("1986881138")
+    expect(Rune._hashFromString("2")).toMatchInlineSnapshot("2072285185")
```
mieszko4, line 274: 🤔 Why do we use `toMatchInlineSnapshot` instead of `toEqual`?

mattbretl: If the result isn't obvious (e.g. the output of this hash function), it's often faster to just write `.toMatchInlineSnapshot()`, let Jest write the missing snapshots, and do a sanity check on the updated files.

mieszko4: Awesome! I did not know Jest can write in test files :)

PR #39, opened 2022-03-03
docs: explain deterministicRandom()

bfelbo:
I find it quite hard to explain the reasoning associated with `deterministicRandom()`. Suggestions welcome! 😄

I'll update the Bunny Twirl example separately once we've released a new SDK version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ Rune provides a random number generator using the challenge number as seed. This
 
 You can use `Rune.deterministicRandom()` instead of `Math.random()` in your map generation code to ensure all players play the same map. The `Rune.deterministicRandom()` function returns a value between 0 and 1 similar to `Math.random()`.
 
-You should only use `Rune.deterministicRandom()` for your map generation and not as a generic replacement for `Math.random()`. This is because each call to `Rune.deterministicRandom()` will iterate through the random values.
+You should only use `Rune.deterministicRandom()` for your map generation and not as a generic replacement for `Math.random()`. This is because each call to `Rune.deterministicRandom()` will iterate through the random values. Any unintentional calls to `Rune.deterministicRandom()` would therefore break the deterministic randomness.
 
 </details>
 
--- a/README.md
+++ b/README.md
@@ -62,7 +62,9 @@ const mapId = mapIds[challengeNumber % mapIds.length] // Get deterministic mapId
 <summary>2) For deterministic randomness (e.g. map generation)</summary>
 &nbsp; 
 
-Rune provides a random number generator seeded with the challenge number and will therefore always provide the same random values for the same challenge number. You can use `Rune.deterministicRandom()` instead of `Math.random()` in your map generation code to ensure all players play the same map.
+Rune provides a random number generator using the challenge number as seed. This random number generator will therefore always provide the same random values for the same challenge number. 
+
+You can use `Rune.deterministicRandom()` instead of `Math.random()` in your map generation code to ensure all players play the same map. The `Rune.deterministicRandom()` function returns a value between 0 and 1 similar to `Math.random()`.
 
 You should only use `Rune.deterministicRandom()` for your map generation and not as a generic replacement for `Math.random()`. This is because each call to `Rune.deterministicRandom()` will iterate through the random values.
 
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ const mapId = mapIds[challengeNumber % mapIds.length] // Get deterministic mapId
 <summary>2) For deterministic randomness (e.g. map generation)</summary>
 &nbsp; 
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -41,23 +41,16 @@ Rune provides a challenge number through `Rune.getChallengeNumber()` that will i
...
-<summary>1) As a seed for randomness</summary>
-&nbsp;
-  
-If your game uses randomness to determine the gameplay (e.g. by randomly generating the maps), you can use the challenge number as the seed to achieve deterministic randomness. For example:
+<summary>1) For deterministic randomness (e.g. map generation)</summary>
+&nbsp; 
 
-```js
-import seedrandom from 'seedrandom' // From https://www.npmjs.com/package/seedrandom
+If your game uses randomness to randomly generate maps, you should use `Rune.deterministicRandom()` instead of `Math.random()`. The random number generator provided by Rune is seeded with the challenge number and will therefore always provide the same random values. In this way, all players will play the same map every time.
```
mattbretl: Just a couple suggestions:
- "uses randomness to randomly generate maps" -> "uses randomness to generate maps"?
- "same random values." -> "same random values. When `Rune.gameOver()` is called, the random number generator is reset with the same seed."

mattbretl: Also, we say "Here's two common ways you could use the challenge number" up above, but for (1), the dev doesn't actually use the challenge number directly. Maybe we should state that under (1) for clarity?

mattbretl:
It might be nice to give one concrete example of 1) and 2) just to drive it home, e.g.
1. A racing game that creates tracks by randomly placing turns, obstacles, and powerups.
2. A racing game that has 20 tracks hand-crafted by the developer.

Also, since 2) is simpler to understand and implement, maybe we should switch the order of 1) and 2)?

README.md
```diff
@@ -41,23 +41,16 @@ Rune provides a challenge number through `Rune.getChallengeNumber()` that will i
...
 
-const challengeNumber = Rune.getChallengeNumber() // Get today's challenge number
-const random = seedrandom(challengeNumber) // Make deterministic number generator
+You should only use `Rune.deterministicRandom()` for your map generation and not as a generic replacement for `Math.random()`. This is because each call to `Rune.deterministicRandom()` will iterate through the random values.
```
mieszko4: Perhaps first we could explain how `Rune.deterministicRandom()` works; that it behaves similar to JS function generator, i.e. every time it is called it fetches the next value of an infinite predefined/fixed sequence of `[0, 1)` numbers:
* the sequence is reset on every game over
* the sequences changes every day based on challengeNumber

And then explain how it can be used for generating elements of game maps with example that @mattbretl has provided.

And then why it is important to not replace all `Math.random()` with `Rune.deterministicRandom()`. A concrete example would be great here, e.g. user-interaction animation.

Also, it is not really clear why a game developer would want to call `Rune.deterministicRandom()` twice so maybe we could provide an example, too.

README.md
bfelbo: Great suggestions, thanks!

> we say "Here's two common ways you could use the challenge number" up above, but for (1), the dev doesn't actually use the challenge number directly. Maybe we should state that under (1) for clarity?

It currently says "The random number generator provided by Rune is seeded with the challenge number", but can try to make it more clear :)

bfelbo:
Great points both of you, thanks a lot!

I've added the racing game examples in https://github.com/rune/rune-games-sdk/pull/39/commits/d4456d89ee9cb1cbe3b32644bb20568ce42775a2. I've also created an internal issue to create an actual new example using deterministic randomness. For now, I'll update based on your feedback so we can get this out there :)

**EDIT:** Clarification.

README.md
bfelbo: I've tried to update this in https://github.com/rune/rune-games-sdk/pull/39/commits/a23b690e7ba5372412602c2502c288632e46aeec and https://github.com/rune/rune-games-sdk/pull/39/commits/8ce2c14a31fe31b043f2b201fd5766a629dbdcb2.

I didn't include "When Rune.gameOver() is called, the random number generator is reset with the same seed." because I think it might be more of a technical detail. I think it's somewhat captured by "This random number generator will therefore always provide the same random values for the same challenge number" and "all players play the same map".

WDYT?

README.md
bfelbo: Great points. I've incorporated some of this in e.g. https://github.com/rune/rune-games-sdk/pull/39/commits/a23b690e7ba5372412602c2502c288632e46aeec, https://github.com/rune/rune-games-sdk/pull/39/commits/8ce2c14a31fe31b043f2b201fd5766a629dbdcb2, and https://github.com/rune/rune-games-sdk/pull/39/commits/1dbfeb82281ee6d5b18752d6c692564a9ac6d0b3.

I've tried to keep it short and simple so haven't included all these points. Similarly, I used a more simplified description than the `[0, 1)` notation in case whoever using the SDK does not have a mathematical background.

bfelbo: See also related comment on https://github.com/rune/rune-games-sdk/pull/39#discussion_r819635387.

bfelbo:
Merging now because we have to get a new SDK version released, but let's keep working on improving this :)

PR #41, opened 2022-03-10
feat: adds invisible overlay in case the game url contains the query param enableInitialOverlayInBrowser=1

Gongreg:
Safari ios throttles requestAnimationFrame when user has not interacted with the iframe at least once.
In case the games are not using clicks (for instance only swiping), ios will not treat these interactions
with the iframe as user interacting. As a workaround, in the browser we will start overlay with
click events disabled and display an invisible div inside the iframe above the canvas.
This way the users will click on the transparent div element the very first time. We will let our client
know about it with _INITIAL_OVERLAY_CLICK event and the transparent div will remove itself.
Afterwards the play/pause will be once again fully controlled by our client.

We also send `_GAME_LOAD_STARTED` event to know when the div is injected.


src/index.ts
```diff
@@ -189,8 +190,45 @@ export const Rune: RuneExport = {
...
+//In case the games are not using clicks (for instance only swiping), ios will not treat these interactions
+//with the iframe as user interacting. As a workaround, in the browser we will start overlay with
+//click events disabled and display an invisible div inside the iframe above the canvas.
+//This way the users will click on the transparent div element the very first time. We will let our client
+//know about it with _INITIAL_OVERLAY_CLICK event and the transparent div will remove itself.
+//Afterwards the play/pause will be once again fully controlled by our client.
+;(() => {
+  const queryParams = Rune._getQueryParams()
+
+  if (!!queryParams.webBrowser && queryParams.webBrowser === '1') {
```
Gongreg: Maybe instead of using webBrowser we could use `enableInitialOverlay`? That would feel a bit more specific.

sanjaypojo (commented):
@Gongreg this looks good to me! Had a question though — what happens in the race condition where the user presses the play button before the iFrame has fully loaded?

Gongreg:
> @Gongreg this looks good to me! Had a question though — what happens in the race condition where the user presses the play button before the iFrame has fully loaded? 

In that case nothing happens (we display play button as inactive).
The iframe should load pretty fast (the initial dom structure without any assets/js). If users clicks play at that moment (while game is not initialized), we will go to "waitingToPlayState". Then as soon as the game fully loads (init called), we will start playing.

src/index.ts
sanjaypojo: 👍🏽 Maybe even `enableInitialOverlayInBrowser` to make both things clear (only enabled in browser and what is being enabled)

sanjaypojo:
> In that case nothing happens (we display play button as inactive). The iframe should load pretty fast (the initial dom structure). If users clicks play at that moment (while game is not initialized), we will go to "waitingToPlayState". Then as soon as the game fully loads (init called), we will start playing.

So to make sure I understood, if I press play before the game loaded, the press will not have an immediate reaction. Then on init, the game starts playing. At this point, is the tap registered correctly and the div removed and iOS perfomance is all good and everything? Otherwise, it might be better to just make them tap again.

Gongreg:
>  if I press play before the game loaded, the press will not have an immediate reaction.

Correct. We will just remember that you clicked play.

> Then on init, the game starts playing.

Yes.

> At this point, is the tap registered correctly and the div removed and iOS perfomance is all good and everything

Yes. Thats the beautiful part.

src/index.ts
```diff
@@ -189,8 +190,45 @@ export const Rune: RuneExport = {
...
+  _getQueryParams: () => {
+    return decodeURI(window.location.search)
+        .replace('?', '')
+        .split('&')
+        .map(param => param.split('='))
+        .reduce((values, [ key, value ]) => {
+          values[ key ] = value
+          return values
+        }, {} as {[key: string]: string})
+  }
```
mieszko4, line 202: Could we use `URLSearchParams` (https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) instead?

src/index.ts
```diff
@@ -189,8 +190,45 @@ export const Rune: RuneExport = {
...
+
+      div.addEventListener('click', () => {
+        div.remove()
+        if (globalThis.postRuneEvent) {
+          globalThis.postRuneEvent({type: "_INITIAL_OVERLAY_CLICK"})
+        }
+      })
+      document.body.appendChild(div)
+    })
+  }
```
mieszko4: > > @Gongreg this looks good to me! Had a question though — what happens in the race condition where the user presses the play button before the iFrame has fully loaded?
> 
> In that case nothing happens (we display play button as inactive). The iframe should load pretty fast (the initial dom structure without any assets/js). If users clicks play at that moment (while game is not initialized), we will go to "waitingToPlayState". Then as soon as the game fully loads (init called), we will start playing.

You mentioned that `we display play button as inactive`. I think that if the play button is disabled user should press on it again after the iFrame loads and the button gets enabled. I am just wondering will it be confusing if I press an inactive button and then after some time the game will just start itself.

Also, in this scenario, iFrame's div will not receive any clicks, so it will not remove the div until next user's click. I.e. user will need to click again anyway to remove the div. And if the game does not require clicks (just swipes) that may seem as broken.

Alternatively, iFrame could remove the div e.g. after game starts, however, I am not sure the first click (while iFrame is not loaded yet) would be considered by the browser as user interaction.

Hence, I would vote for enabling Play button only after receiving `INIT` from iFrame.

src/index.ts
```diff
@@ -203,3 +241,4 @@ export type RuneGameEvent =
   | { type: "GAME_OVER"; score: number; challengeNumber: number }
   | { type: "ERR"; errMsg: string }
   | { type: "SCORE"; score: number; challengeNumber: number }
+  | { type: "_INITIAL_OVERLAY_CLICK"}
```
mieszko4, line 248: I do not think we have to prefix that with `_`. Could this be simply `INITAL_OVERLAY_CLICK`?

mieszko4 (commented):
Nice workaround. I have some questions before we merge this :)

src/index.ts
Gongreg: I didn't want to use it to be as compatible with as many browser versions as possible (since this is a pretty simple logic).

src/index.ts
Gongreg: This is an internal event that sdk users don't need to worry about. using _ indicates that :)

src/index.ts
Gongreg: I did it this way to replicate the way our app behaves. 
The next improvement will be to have a loading animation (like in the app).
 After users clicks on the iframe (which is identical to clicking play button in the app), we will hide the play button (only leave the loading animation) in our client to indicate that the user is waiting for the load to finish.

Gongreg: And I guess we should not display the button as inactive while the games is loading, true.
Will create an issue for it.

mieszko4 (approved):
🎉

src/index.ts
bfelbo: I'm not sure this event is more "internal" than the other events? None of these events are used by game devs. They're only sent to the Rune app / browser :)

Gongreg: Ah, I see. We can change it to not use _ then. 
The only thing is that they are "hackish" events, so wondering how to indicate that.

bfelbo:
The tests seem to fail after merging this. Let's use `globalThis` instead of `window`?

```
 FAIL  test/sdk.test.ts
  sdk
    ✕ init() -> startGame() -> pauseGame() -> resumeGame() (6 ms)
    ✕ don't allow calling other functions before init() (6 ms)
    ✕ ensure correct properties passed to init() (2 ms)
    ✕ ensure correct types passed to init() (2 ms)
    ✕ ensure score passed as number (2 ms)
    ✕ exposed version should match npm version (2 ms)
    ✕ INIT event should include version matching npm version (2 ms)
    ✕ SCORE event should include score from game's getScore() and challenge number (1 ms)
    ✕ GAME_OVER event should include score from game's getScore() and challenge number (1 ms)
    ✕ allow replacing functions through code injection (2 ms)
    ✕ get challenge number if not injected (1 ms)
    ✕ get challenge number if injected (1 ms)
    ✕ deterministicRandom() works before init() (1 ms)
    ✕ deterministicRandom() changes value based on challengeNumber (1 ms)
    ✕ deterministicRandom() resets with gameOver() (1 ms)
    ✕ deterministicRandom() seed changes dramatically with every challenge number (2 ms)

  ● sdk › init() -> startGame() -> pauseGame() -> resumeGame()

    The error below may be caused by using the wrong test environment, see https://jestjs.io/docs/en/configuration#testenvironment-string.
    Consider using the "jsdom" test environment.
    
    ReferenceError: window is not defined

      192 |   },
      193 |   _getQueryParams: () => {
    > 194 |     return decodeURI(window.location.search)
          |                      ^
      195 |       .replace("?", "")
      196 |       .split("&")
      197 |       .map((param) => param.split("="))

      at Object._getQueryParams (src/index.ts:194:22)
      at src/index.ts:213:28
      at Object.<anonymous> (src/index.ts:239:3)
      at test/sdk.test.ts:12:11

  ● sdk › init() -> startGame() -> pauseGame() -> resumeGame()

...
    
 › 4 snapshots failed.
Snapshot Summary
 › 4 snapshots failed from 1 test suite. Inspect your code changes or run `npm test -- -u` to update them.

Test Suites: 1 failed, 1 total
Tests:       16 failed, 16 total
Snapshots:   4 failed, 4 total
Time:        0.753 s, estimated 3 s
Ran all test suites.
```

src/index.ts
bfelbo: Makes sense. Another way could be `BROWSER_INITIAL_OVERLAY_CLICK` and `BROWSER_GAME_LOAD_STARTED`. That way, any devs working on the Rune app would know that it's not relevant to them.

Anyway, mostly just wanted to clarify any potential confusion about who/what uses the events :)

bfelbo: If we wanna change it, we should probably do it before rolling out to production to keep our life simple.

bfelbo: FYI @mattbretl in case you have any thoughts on whether wanna use _ for events going forward :)

mattbretl: The `BROWSER_` prefix sounds great to me. 👍

Gongreg: I'll change it to BROWSER before releasing to production, thanks guys :)

PR #45, opened 2022-03-29
chore: add rollup to support multiple files

Gongreg:
* Adding Rollup bundler to support using multiple files.
To generate output similar to currently existing one, I am using IIFE format for browser with Rune global variable and CommonJs for index.js (I think we could instead op to ES format).


dist/browser.js
```diff
@@ -1,194 +1,213 @@
...
-        // When running inside Rune, runePostMessage will always be defined.
-        if (window.postRuneEvent) {
-            window.postRuneEvent({ type: "INIT", version: Rune.version });
-        }
-        else {
-            Rune._mockEvents();
-        }
-    },
-    gameOver: function () {
+var Rune = (function () {
```
bfelbo, line 1: `var Rune` seems to be defined both here and below? Might be good to only define it once? :)

dist/browser.js
```diff
@@ -1,194 +1,213 @@
-/*
```
bfelbo, line 1: If easy, it'd be nice to keep the same order as before to minimize the diff and keep `var Rune = { version: ... }` at the top.

src/index.ts
```diff
@@ -189,58 +189,10 @@ export const Rune: RuneExport = {
...
-      })
-      document.body.appendChild(div)
-
-      if (globalThis.postRuneEvent) {
-        globalThis.postRuneEvent({ type: "BROWSER_IFRAME_LOADED" })
-      }
-    })
-  }
-})()
+setup();
```
bfelbo: Would be nice with a comment explaining what this is w/o having to open the file :)

src/index.ts
```diff
@@ -1,6 +1,7 @@
 /*
 The SDK interface for games to interact with Rune.
 */
+import {setup} from './internal';
```
bfelbo: We have other internal functions (see e.g. `_randomNumberGenerator` below) that we may wanna move out into a separate file. Perhaps it'd be cleaner to name this file as `browser`, `internal/browser`, or similar? :)

src/index.ts
bfelbo: Might not be needed if we rename the file/functions.

bfelbo (commented):
Approach looks good to me 😄

Would be great if you could test that the `browser.js` file still works. The easiest way is to copy it locally into the examples folder and make the example use the local file instead.

As we add more bundling complexity it'd be awesome to ensure that we still release ES3-compliant code. If it's easy, it would be nice to add a test for this before merging this. Perhaps using https://github.com/yowainwright/es-check or similar. If it's hard, then we can do it manually for now :)

dist/browser.js
Gongreg: Discussed in a voice chat. Since everything is inside a function, the naming of variables inside does not matter.

dist/browser.js
Gongreg: Discussed in the voice chat. The order does not matter too much, since we expect users to read the api/source code and not the dist output.

src/index.ts
Gongreg: I've renamed the file/function to `setupBrowser` for now.

src/index.ts
Gongreg: Renamed to `setupBrowser`.

Gongreg:
Tested out the new bundle, the example game is working correctly with it.

Added a command to test output: `test:browserCompatibility`. Also added execution of it inside `test`. After we have a CI running for eslint/prettier etc. we can move it out from `test` so it does not run locally.

src/setupBrowser.ts
```diff
@@ -0,0 +1,50 @@
...
+    }, {} as { [key: string]: string })
+}
+
+export function setupBrowser() {
+  //Safari ios throttles requestAnimationFrame when user has not interacted with the iframe at least once.
+  //In case the games are not using clicks (for instance only swiping), ios will not treat these interactions
+  //with the iframe as user interacting. As a workaround, in the browser we will start overlay with
+  //click events disabled and display an invisible div inside the iframe above the canvas.
+  //This way the users will click on the transparent div element the very first time. We will let our client
+  //know about it with _INITIAL_OVERLAY_CLICK event and the transparent div will remove itself.
```
mieszko4: Nit:
```suggestion
  //know about it with BROWSER_INITIAL_OVERLAY_CLICKED event and the transparent div will remove itself.
```

bfelbo (approved):

PR #46, opened 2022-03-29
fix: disable local storage

Gongreg:
Some of the games using Rune SDK use local storage internally, which is breaking our usage.
Added logic to disable the local storage + logging an error in case the game is using it. The method 'getItem' is always returning null to simulate empty localStorage and not break the games depending on local storage.

I've tested it out on one of the real games, this solution was working & fixed the issues related to local storage usage.
We'll still need to test the other games to see if something isn't breaking due to this change.


bfelbo:
Amazing that this works and that you've tested it with a game. 

@mieszko4 would be awesome if you could take a look :)

src/setupBrowser.ts
```diff
@@ -48,3 +48,39 @@ export function setupBrowser() {
     })
   }
 }
+
+function disableLocalStorage() {
```
bfelbo: I think we also need to disable it for games running inside our app? If so, we might want to move it into another file for clarify :)

Gongreg: right. My bad.

Gongreg: Moved `disableLocalStorage` to separate file & moved `disableLocalStorage`/`setupBrowser` files inside `internal` folder to not clutter the entry dir.

src/internal/disableLocalStorage.ts
```diff
@@ -0,0 +1,30 @@
...
+    },
+    setItem: {
+      value: noop,
+    },
+    removeItem: {
+      value: noop,
+    },
+    clear: {
+      value: noop,
+    },
```
mieszko4: Based on https://developer.mozilla.org/en-US/docs/Web/API/Storage there are 2 more items in `localStorage`:
* `.length` read-only property - we should return 0
* `.key(index: number)` method - we should return `null`

src/internal/disableLocalStorage.ts
```diff
@@ -0,0 +1,30 @@
+export function disableLocalStorage() {
+  if (!globalThis.localStorage) return
```
mieszko4: Should we also disable `sessionStorage`?

Since both `localStorage` and `sessionStorage` inherit from `Storage` then we could make generic `disableStorage(storage: Storage)` function and use it as:
```ts
disableStorage(globalThis.localStorage)
disableStorage(globalThis.sessionStorage)
```

Gongreg: Both storages use the same Storage object, so this is actually disabling both (because I am changing the prototype). I'll just rename the method to disableStorage.

src/internal/disableLocalStorage.ts
Gongreg: alright. I've never seen anyone use them though :D .

src/internal/disableLocalStorage.ts
Gongreg: Changed in [`4a97ce8` (#46)](https://github.com/rune/rune-games-sdk/pull/46/commits/4a97ce81f80920252be0c0beed67a286d7ba4839)

src/internal/disableLocalStorage.ts
Gongreg: Changed in [`4a97ce8` (#46)](https://github.com/rune/rune-games-sdk/pull/46/commits/4a97ce81f80920252be0c0beed67a286d7ba4839)

src/internal/disableLocalStorage.ts
mieszko4: > Both storages use the same Storage object, so this is actually disabling both (because I am changing the prototype). I'll just rename the method to disableStorage.

Good point! Would be great to add comment in the code where you `Object.defineProperties` because it is not clear from that this will disable both localStorage and sessionStorage.

src/internal/disableLocalStorage.ts
mieszko4: 👍 Just a minor thing: accessing `.length` does not return a warning - but I guess that's fine to keep it simple. Otherwise we would need to use `Proxy` :)

src/internal/disableLocalStorage.ts
Gongreg: Added a comment explaining it.

src/internal/disableLocalStorage.ts
Gongreg: yes, and proxy is not very backwards compatible friendly.

src/internal/disableStorage.ts
```diff
@@ -0,0 +1,36 @@
+export function disableStorage() {
+  if (!globalThis.localStorage) return
+
+  const noop = () => {
+    console.error("WARNING! Local storage is disabled when using Rune SDK.")
```
mieszko4: FFTI: You use `console.error` but you prefix the msg with `WARNING` -  so that seems a bit inconsistent.
Also maybe you could improve the msg by adding that both local and session storages are disabled.

mieszko4 (approved):
Looks great!

src/internal/disableStorage.ts
Gongreg: I am notifying user not to use the storage, but it does not break the app, so I sent the warning message. But the yellow message is a bit easier to miss, especially if user is using some kind of game engine etc where getting some warnings in the console log is expected. I'll replace the text to ERROR since in the end we don't want the user to use the storage.

Gongreg: [`ae4828c` (#46)](https://github.com/rune/rune-games-sdk/pull/46/commits/ae4828c47848de4f45534242dd4e9a5ef3f4797c)

helios1138:
@Gongreg to our yesterday's conversation, this change would be another reason why we can't embed the games into the web app without an iframe 🙂

Gongreg:
This case could be worked around differently (for instance allowing only the keys that we use ourselves to be saved to storage). But yes, it would complicate this and many other things most likely.

PR #47, opened 2022-03-31
fix: clear storage instead of disabling it

Gongreg:
Some of the games actually use the local storage not only as a backup but as a place to store the data.

This PR removes storage disabling and instead calls .clear() before game load.


helios1138:
@Gongreg 
> An alternative could be to call original methods in our overrides & call localStorage/sessionStorage clear method at the load.

Perhaps we can just not override anything and call `.clear()` at the start/init?

Gongreg:
That's what I wrote. But we still want to notify to show error so devs would know not to use local storage moving forward.
I could use a proxy for that, but I am not sure that all the browsers that users use will support it.

Gongreg:
Clear method would work consistently, since our sdk code is inlined and should execute before the games is loaded.

So these are the options:
1. Just call `localStorage/sessionStorage.clear`. Simplest solution, but no logs in console.
2. call `localStorage/sessionStorage.clear` + Add Proxy for local/session storage. clean solution/logs in console. Compatibility: https://caniuse.com/?search=proxy
3. call `localStorage/sessionStorage.clear`  + override  `localStorage/sessionStorage` to call console log but still call original methods.
4. fully override local/session storage with local object.

If we are okay with Proxy support I would vote for 2. Otherwise I think we could do the option 3.

helios1138:
@Gongreg we can also just check the game code for localStorage usage during addGame/updateAllGames scripts and have a warning displayed to us and then we can relay it to the developers. (When adding games will be done using an API by developers themselves, we could show the warning to them directly). It seems that maybe just telling them this directly is better than handicapping the already written code in this way (perhaps there will be more unintended consequences?).

Gongreg:
That seems like a pretty good approach. not sure we will be able to very consistently notify about the usage in the API, but that is for the future. For now I could change this PR to just call `clear` and that's it.

bfelbo:
Great discussion. It would be nice for devs to get warnings as they code and not on submission. Sounds like it's hard to do that in a clean way so sounds good to just call `clear` 👍

Gongreg:
@bfelbo - I could change this PR to call existing methods + clear at the start. Wouldn't make things complicated.

helios1138:
@bfelbo to me it just doesn't look nice to simply warn them in the console. If we definitely 100% do not want to allow localStorage then it should be an error/crash/something drastic. Otherwise, it's a bit unclear if they can use it, but there will be warnings or if they will be using it assuming one behavior but we override its behavior to be different and so on.

helios1138:
@bfelbo I don't know how extensive is our current level of cooperation with developers, ideally, we should just define the conditions/restrictions we want and have them adapt. Because if they are using localStorage assuming long-term data recovery perhaps at some fundamental level they simply are not producing the games/experience we expect?

Gongreg:
@helios1138 , I've updated the PR to use only `.clear()`.

helios1138 (approved):

PR #53, opened 2022-05-04
feat: move native app message bridge into the sdk

Gongreg:
## Sdk:
* Added the postMessage communication support for native app.
* Exposed `getRuneGameEvent, stringifyRuneGameCommand` methods that we can use in clients to create communication messages.
* Using `runeGameCommand` key for commands comming from postMessages into the RuneSDK & `runeGameEvent` for sending events back to the clients.
* Added basic validations for the postMessage data.
* Stopped using globalThis for postMessage/challengeNumber while keeping legacy client support for globalThsi challengeNumber/postMessage

## Tests:

* Added tests for post message communication.
* Rune sdk accessible with get function to avoid cluttering global context.
* Test watch command.
* Removed --runInBand flag since now the tests are not cluttering global scope.

Tested scenarios:

- [x] Legacy native app.
- [x] Updated native app.
- [x] Updated web.


bfelbo:
Sounds great to standardize it and add tests 💯

@mattbretl would be great if you could take a look!

bfelbo:
Sounds great to standardize it and add tests 💯

@mattbretl would be great if you could take a look!

src/messageBridge.ts
```diff
@@ -0,0 +1,73 @@
+import { RuneGameCommand, RuneGameEvent } from "./types"
+
+const RUNE_MESSAGE_PREFIX = "RUNE_MSG;"
```
bfelbo, line 6: Might be nice with a comment to explain the rationale behind using a prefix before the stringified JSON :)

src/internal/setupMessageBridge.ts
```diff
@@ -0,0 +1,36 @@
...
+import { RuneGameEvent } from "../types"
+
+export function setupMessageBridge() {
+  globalThis.postRuneEvent = (data: RuneGameEvent) => {
+    //We only expect to send Game events through postMessages
+    const event = stringifyRuneGameEvent(data)
+
+    globalThis.ReactNativeWebView
+      ? //Post message for Native app
+        globalThis.ReactNativeWebView.postMessage(event)
```
helios1138: oh, that's why it wasn't working before!

src/internal/setupMessageBridge.ts
```diff
@@ -0,0 +1,36 @@
...
+      //Ignore non Rune messages
+      if (!command) {
+        return
+      }
+
+      if (!command.type || !globalThis.Rune[command.type]) {
+        throw new Error(`Received incorrect message!: ${command}`)
+      }
+
+      globalThis.Rune[command.type]()
```
helios1138: perhaps now that we're refactoring this, it might be a good time to add some `arguments` to the command, in case in the future we'll want to use them. To add `args?: {}` to `RuneGameCommand` and pass them into callback/method here.

src/messageBridge.ts
Gongreg: Thanks, good point! Added the comment: [`c73e270` (#53)](https://github.com/rune/rune-games-sdk/pull/53/commits/c73e270eda6ce3d9c97c7971b22de94c921134c3)

src/internal/setupMessageBridge.ts
Gongreg: As discussed in voice chat, we are not adding it for now.

src/types.ts
```diff
@@ -0,0 +1,68 @@
...
+export interface RuneExport {
+  // External properties and functions
+  version: string
+  init: (input: InitInput) => void
+  gameOver: () => void
+  getChallengeNumber: () => number
+  deterministicRandom: () => number
+
+  // Internal properties and functions
+  _doneInit: boolean
```
mattbretl, line 18: FFTI: Instead of this, should we just declare a variable, e.g. `let _doneInit: boolean = false` to align with `doneFirstPlay`?

mattbretl (approved):
LGTM!

src/types.ts
Gongreg: We'll be moving it out in another PR which will be responsible for event order validation. I tried not to touch the Rune object itself in this PR.

PR #55, opened 2022-05-06
fix: add support for android webview post messages

Gongreg:
According to https://github.com/react-native-webview/react-native-webview/issues/356  to support postMessages on android React Native webview, we need to listen to document events.


src/internal/getUrlParams.ts
```diff
@@ -0,0 +1,9 @@
+export function getUrlParams() {
```
bfelbo, line 1: Would it make sense to add types here? :)

bfelbo (approved):

PR #56, opened 2022-05-09
fix: legacy client challenge number

Gongreg:
There seems to be a race condition for game challenge. Added recheck for legacy flow on each read to avoid it.


bfelbo (approved):
Looks great 🥳

We can also move / remove [these tests](https://github.com/rune/rune-games-sdk/blob/staging/test/sdk.test.ts#L191-L204).

mieszko4 (approved):
👍

PR #58, opened 2022-05-23
refactor: use state machine

Gongreg:
Added state machine  https://stately.ai/registry/editor/share/ced5de88-385e-44f3-938f-ffa38580b774 .

Removed `startGame` callback.
Added `restartGame` callback.

Added error state which is entered in case something unexpected happens (multiple init/game over calls).


test/sdk.test.ts
```diff
@@ -2,80 +2,50 @@ import {
...
-  test("don't allow calling other functions before init()", async function () {
-    expect(
-      await extractErrMsg(() => {
-        Rune._startGame()
-      })
-    ).toMatchInlineSnapshot(`"Rune._startGame() called before Rune.init()"`)
-  })
-
   test("ensure correct properties passed to init()", async function () {
+    const { Rune } = getRuneSdk(1)
```
bfelbo: Could consider also using `CHALLENGE_NUMBER` here to avoid magic constant `1` :)

bfelbo (approved):
Changes look great! 🥳 Mentioned a few suggestions when we spoke. Feel free to merge with or w/o them.

PR #61, opened 2022-05-25
fix: add support for updated client with old games

Gongreg:
Changes to support games which are not updated yet to the new sdk on updated client.


bfelbo (approved):
💯

PR #73, opened 2022-06-07
docs: add quick start and improve organization

sanjaypojo:
- Tried to break down the paragraphs into bullet points
- Added a section for Rune.gameOver() — now the sections match the API and `Rune.gameOver()` is harder to miss.

I will do a separate PR for the challenge number section.


bfelbo (approved):
🥳

PR #84, opened 2022-06-28
feat: alternative way of getting the score

helios1138:
Added support for getting the score using a field on `getScore` function for environments where returning from a callback is not supported.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -11,6 +11,15 @@ export function getRuneSdk({ challengeNumber }: { challengeNumber: number }) {
   const Rune: RuneExport = {
     version: "2.2.3",
     init: (input) => {
+      if (input.getScore.canNotReturn) {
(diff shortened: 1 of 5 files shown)


src/index.ts
```diff
@@ -11,6 +11,15 @@ export function getRuneSdk({ challengeNumber }: { challengeNumber: number }) {
...
     version: "2.2.3",
     init: (input) => {
+      if (input?.getScore?.canNotReturn) {
+        const originalGetScore = input.getScore
+
+        input.getScore = () => {
+          originalGetScore()
+          return originalGetScore.score
+        }
+      }
```
helios1138: this is the actual new code, the rest of the PR are tests, types and `prettier` changes

bfelbo (approved):
Awesome how you managed to make this workaround quite clean! 💯

PR #99, opened 2022-10-13
feat: add viewport meta tag to html5 game example

helios1138:


helios1138:
@bfelbo merging w/out review

PR #108, opened 2022-12-16
feat: allow Math.random

oblador:
Next version of the Multiplayer SDK supports randomness, so this PR updates the linter to allow it and adds a section in the docs about it.

### DIFF
--- a/docs/docs/multiplayer/randomness.md
+++ b/docs/docs/multiplayer/randomness.md
@@ -21,9 +21,9 @@ Rune solves this problem by implementing a deterministic pseudorandom number gen
 
 What's great about it is that the game can do optimistic updates which leads to a really fast game, and that cheating is hard as the player client only knows how to generate their own numbers and cannot know what the other players will get.
 
-## Great, but sounds complicated?
+## Things to keep in mind
 
-Yes, but not for your game! Rune solves this transparently and in the majority of cases **you can keep using `Math.random()`** as you would in any other game. There are some specific things to keep in mind:
+Rune solves this transparently and in the majority of cases **you can keep using `Math.random()`** as you would in any other game. There are some specific things to keep in mind however:
 
 ### Keep all shared state in `logic.js`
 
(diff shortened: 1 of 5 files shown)


docs/docs/multiplayer/randomness.md
```diff
@@ -0,0 +1,28 @@
+---
+sidebar_position: 6
+---
+
+# Randomness
+
+Most games involve some kind of randomness, for example to roll a die in the case of Yahtzee. However randomness introduces some problems:
```
bfelbo:
```suggestion
Most games involve some kind of randomness, for example to roll a die in the case of Yahtzee. However, randomness introduces some problems:
```

docs/docs/multiplayer/randomness.md
```diff
@@ -0,0 +1,28 @@
...
+
+- True randomness is inherently **not deterministic**, which is not compatible with Rune's distributed game state model.
+- True randomness with optimistic updates places trust in the client, and therefore **opens up to cheating** for a technically sophisticated player.
+
+## Deterministic random
+
+Rune solve this problem by implementing a deterministic pseudorandom number generator. In simpler words this means the Rune server tells each player client to calculate random numbers. For every action the player takes, the server verifies that the random numbers being generated are in fact generated in the way the server told it to.
```
bfelbo:
```suggestion
Rune solves this problem by implementing a deterministic pseudorandom number generator. In simpler words this means the Rune server tells each player client to calculate random numbers. For every action the player takes, the server verifies that the random numbers being generated are in fact generated in the way the server told it to.
```

docs/docs/multiplayer/randomness.md
```diff
@@ -0,0 +1,28 @@
...
+
+Yes, but not for your game! Rune solves this transparently and in the majority of cases **you can keep using `Math.random()`** as you would in any other game. There are some specific gotchas that is good to know about:
+
+### Keep all shared state in `logic.js`
+
+This isn't limited to randomness, but is especially important here, as only the code called from within `Rune.initLogic()`/`logic.js` has the deterministic random. If you have some kind of initialization code that generates data for your game, make sure to do this in `setup()` so that all players see the same thing.
+
+### Shared random
+
+If your games needs a deterministic shared randomness, meaning that the randomness is determined by the order an action is taken, regardless of who makes an action. This can be solved in different ways depending on the needs. One way is to use `Math.random()` in `setup()` to set a shared random state to be used in your own number generator, another is to generate all state that depends on randomness in `setup()`.
```
bfelbo:
```suggestion
If your games needs deterministic shared randomness, meaning that the randomness is determined by the order an action is taken, regardless of who makes an action. This can be solved in different ways depending on the needs. One way is to use `Math.random()` in `setup()` to set a shared random state to be used in your own number generator, another is to generate all state that depends on randomness in `setup()`.
```

bfelbo (approved):
Awesome to have randomness! 🥳 Looks great. Suggested some minor typo fixes and have a few thoughts on changing the docs as discussed elsewhere.

PR #109, opened 2022-12-16
Release packages with lerna

oblador:
This PR introduces lerna to manage versions and releasing packages with GitHub actions.


bfelbo (approved):
LGTM!

PR #118, opened 2023-01-09
build: manually manage package versions

oblador:


helios1138:
@oblador what does the flow looks like now?

PR #124, opened 2023-01-21
docs: invalidAction is no longer a function

bfelbo:
Tiny change.


oblador:
Should probably also update the multiplayer example

bfelbo:
Great point. You wanna do it and merge?

oblador:
Well, actually I think we should reconsider this change. I don't really see the gain in this breaking API change, however I do see a downside in that stack trace will be pointing to the wrong place which would harm certain debugging cases in the future.

PR #126, opened 2023-03-03
docs: avatar placeholder assets

oblador:
This PR adds a section in the docs explaining the need for placeholder images and provides assets for devs to use:

Screenshot 2023-03-03 at 17 36 06


docs/docs/multiplayer/player-info.md
```diff
@@ -16,6 +16,14 @@ Here’s the values for each player:
 - `avatarUrl: string`
 - `playerId: string` (same as key, just provided for simplicity)
 
+## Avatars
+
+Using avatars is a great way to personalize the UI to show who's turn it is or in a leaderboard. Since the avatar is loaded over the network there might be a slight delay during which you might want to display a placeholder – and we got you covered!
```
mieszko4, line 21: Should this be `whose` instead of `who's`?

mieszko4 (approved):
👍

docs/docs/multiplayer/player-info.md
oblador: Nope, it's short for `who is`

mieszko4: But then you are saying `to show who is turn it is` instead of `to show whose turn it is`. That sounds off.

oblador: You're right! Fixed in https://github.com/rune/rune-games-sdk/commit/bd7c4475b342c7d84c633fa5001a4e5555555ae3

PR #129, opened 2023-03-13
feat: add gameOver() to tic-tac-toe example

helios1138:

### DIFF
--- a/multiplayer/examples/tic-tac-toe/logic.js
+++ b/multiplayer/examples/tic-tac-toe/logic.js
@@ -30,6 +30,7 @@ Rune.initLogic({
     lastPlayerId: null,
     cells: new Array(9).fill(null),
     winCombo: null,
+    gameOver: false,
   }),
   actions: {
     claimCell: (cellIndex, { game, playerId }) => {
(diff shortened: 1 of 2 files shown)


multiplayer/examples/tic-tac-toe/logic.js
```diff
@@ -47,7 +48,31 @@ Rune.initLogic({
...
+          Rune.gameOver({
+            players: {
+              [winner]: "WON",
+              [loser]: "LOST",
+            },
+            delayPopUp: true,
+          })
+        } else {
+          // Game is a draw
+          Rune.gameOver({
```
oblador, line 68: Interesting, I guess you can't just do a generic game over if you also want to use `delayPopUp: true`

oblador (approved):

multiplayer/examples/tic-tac-toe/logic.js
helios1138: It's possible to do only without `delayPopUp: true`, yes. But I also thought it would actually make sense to have "everybody lost" as the "draw" result of a 2-player game

PR #130, opened 2023-03-17
feat: improve game files validation

helios1138:

### DIFF
--- a/packages/rune-games-cli/src/lib/validateGameFiles.ts
+++ b/packages/rune-games-cli/src/lib/validateGameFiles.ts
@@ -4,14 +4,14 @@ import path from "path"
 import semver from "semver"
 
 import { extractMultiplayerMetadata } from "./extractMultiplayerMetadata.js"
-import { FileInfo } from "./getGameFiles.js"
+import { FileInfo, findShortestPathFileThatEndsWith } from "./getGameFiles.js"
 import { rootPath } from "./rootPath.js"
 
 import LintMessage = Linter.LintMessage
 
 export const validationOptions = {
   sdkUrlStart: "https://cdn.jsdelivr.net/npm/rune-games-sdk",
-  sdkVersionRegex: /rune-games-sdk@(\d\.\d(\.\d)?)/,
+  sdkVersionRegex: /rune-games-sdk@(\d(\.\d(\.\d)?)?)/,
   minSdkVersion: "4.5.0",
   maxFiles: 1000,
   maxSizeMb: 25,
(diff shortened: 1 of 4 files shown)


packages/rune-games-cli/src/lib/validateGameFiles.ts
```diff
@@ -112,7 +108,14 @@ export async function validateGameFiles(
         .getAttribute("src")
         ?.match(sdkVersionRegex)?.[1]
 
-      const sdkVersionCoerced = semver.coerce(sdkVersion)
+      if (!sdkVersion) {
+        errors.push({ message: `Rune SDK must specify a version` })
+      }
+
+      const [major, minor, patch] = (sdkVersion ?? "")?.split(".")
+      const maxedOutSdkVersion = `${major}.${minor ?? 999}.${patch ?? 999}`
```
Gongreg, line 116: why do we do 999?

packages/rune-games-cli/src/lib/validateGameFiles.ts
```diff
@@ -61,12 +61,8 @@ export async function validateGameFiles(
     errors.push({ message: `Game size must be less than ${maxSizeMb}MB` })
   }
 
-  const indexHtml = files.find(
-    (file) => file.path === "index.html" || file.path.endsWith("/index.html")
-  )
-  const logicJs = files.find(
-    (file) => file.path === "logic.js" || file.path.endsWith("/logic.js")
-  )
+  const indexHtml = findShortestPathFileThatEndsWith(files, "index.html")
```
Gongreg, line 64: This does work, but can't we just do root/index.html, root/logic.js instead of searching/sorting?

bfelbo: I think it's fine to enforce that the files are in root 👍

packages/rune-games-cli/src/lib/validateGameFiles.ts
helios1138: to allow comparing e.g. `sdk@4` with `minVersion: 4.5.0`. semver package resolves `4` to `4.0.0` but it's actually "the latest 4.x.x" version

Gongreg: I see, missed the variable name, thanks :)

packages/rune-games-cli/src/lib/validateGameFiles.ts
helios1138: Because validation is reused in cli, tango and also exists as a tango API endpoint, changing everything to now be aware of the root would require more work. maybe we can do this later, for now I just want to fix this pressing issue in a quickest way

helios1138:
Gonna merge w/out review

PR #134, opened 2023-04-13
fix: binary expression of object properties not allowed

oblador:
Fixes that the following (safe code) wasn't allowed:

```js
const guesses = game.guesses.filter(
  (guess) => guess.round === game.currentRound
)
```


helios1138 (approved):
👌

PR #136, opened 2023-04-20
feat: allow parent scope reads but not writes

oblador:
This PR loosens the restrictions on accessing parent scope variables to allow any kind of read. We didn't allow this before as it can be bypassed by passing it as an argument to a function that then mutates its argument (as tests show) however this is likely such a rare case that we can ignore it for now. 

It also allows assigning globals to variables, but still protects them from mutation (within limits, should at least catch accidental).

Additionally it allows `var`'s but as a warning.


bfelbo (approved):
LGTM!

helios1138 (approved):
LGTM 👍

PR #143, opened 2023-05-10
feat: allow export statements in logic.js

oblador:
This PR updates the linting config to allow safe export statements for code re-use with client end code.

### DIFF
--- a/packages/eslint-plugin-rune/src/index.ts
+++ b/packages/eslint-plugin-rune/src/index.ts
@@ -130,7 +130,7 @@ const logicConfig: ESLint.ConfigData = {
       "error",
       {
         selector:
-          "ImportDeclaration,ExportNamedDeclaration,ExportAllDeclaration,ExportDefaultDeclaration",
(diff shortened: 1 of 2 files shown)


helios1138:
@oblador How do we then guarantee that there won't be exports in the final bundle for upload? Or are we actually ok with this?

oblador:
Why would an export of a local variable be problematic?

helios1138:
Also, AFAIK importing from logic.js inevitably means that initLogic() would be called twice now.

helios1138:
> Why would an export of a local variable be problematic?

Good question. Why did we prohibit it in the first place?

oblador:
Because `export ... from` is problematic and I didn't think of any good other uses for export (but I do now 😇)

oblador:
> Also, AFAIK importing from logic.js inevitably means that initLogic() would be called twice now.

If your client code imports from logic.js you don't have to add it as a script tag, but even if you did all other module systems that I know makes sure that a file is only evaluated once.

helios1138 (approved):

bfelbo (approved):
🥳

PR #144, opened 2023-05-11
Add `vite-plugin-rune` package and vite template

oblador:


multiplayer/vite-template/src/logic.ts
helios1138, line 1: Can we add some examples of shared imports here like we discussed? E.g. some lib function that is imported both in the logic.ts and also somewhere in the client.

oblador: I don't want to do that in the template, bur rather as a test of the plugin itself.

oblador: @helios1138 I've added some integration tests now 👍

helios1138 (commented):
Have a couple of questions:
- I tried to test it, but I get a bunch of type errors everywhere. I'm guessing it has to do with multiple "typescripts" inside the monorepo. Not sure if we can do something about it.
- `yarn dev` doesn't start the game for me (`caught (in promise) Error: sdk init failed`)
- why does `main.tsx` imports `App.tsx` as `"./App.tsx"`, but for example `App.tsx` imports `logic.ts` as just `"./logic"`?
- why does `index.html` not include the SDK script?
- why does `index.html` include `logic.ts` script?
- should `vite-template` go under `multiplayer/examples` rather than `multiplayer/`?

helios1138 (approved):

PR #146, opened 2023-05-18
feat: allow not importing logic.js as a script tag

oblador:
Since we now allow exports in the logic file as of #143 along with other changes such as supporting vite, and our vite-template actually not injecting it, it no longer makes sense to require it to be included via a script tag as it might just as well be imported from the client.js itself. 

### Before

Screenshot 2023-05-18 at 10 26 52

### After

No errors

### DIFF
--- a/packages/rune-games-cli/src/lib/validateGameFiles.ts
+++ b/packages/rune-games-cli/src/lib/validateGameFiles.ts
@@ -1,5 +1,6 @@
 import { ESLint, Linter } from "eslint"
 import { parse, valid } from "node-html-parser"
+import path from "path"
 import semver from "semver"
 
 import { extractMultiplayerMetadata } from "./extractMultiplayerMetadata.js"
@@ -95,6 +96,7 @@ export async function validateGameFiles(
           multiplayerValidationResult,
           errors,
           logicJs,
+          indexHtml,
         })
       }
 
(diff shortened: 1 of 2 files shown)


bfelbo:
Sounds good to me. Will let @helios1138 take the lead on review :)

packages/rune-games-cli/src/lib/validateGameFiles.ts
```diff
@@ -142,51 +139,23 @@ export async function validateGameFiles(
...
-    }
-
-    if (!logicJs) {
-      errors.push({
-        message: "logic.js must be included in the game files",
-      })
-    } else {
-      if (path.dirname(indexHtml.path) !== path.dirname(logicJs.path)) {
-        errors.push({
-          message: "logic.js must be in the same directory as index.html",
```
helios1138, line 180: I think we want to keep this check

oblador: Added it back in [9f0b98f](https://github.com/rune/rune-games-sdk/pull/146/commits/9f0b98f05761d238437819e67f0f8c2c8e8dfdad)

helios1138 (approved):
LGTM!

helios1138 (approved):
LGTM!

helios1138 (approved):
LGTM!

helios1138 (approved):
LGTM!

PR #149, opened 2023-05-19
Outmatched example

oblador:
This PR adds the example game Outmatched.

It could use a refactor as this was originally written with constraints of a single logic/client file, but has since been refactored to TS and to use Vite, but I deemed it better to have something to iterate on than nothing. 

Screenshot 2023-05-19 at 16 03 59


multiplayer/examples/outmatched/src/sounds/index.ts
```diff
@@ -0,0 +1,3 @@
+export const playSound = (_name: string) => {
+  // We don't have the license to freely redistribute sounds used in this game
```
bfelbo: Would it make sense to keep the playSound code in the repo, but just fail silently and print to console if the sound file is not there?

It'd be great to have the exact same code in the example folder as used to upload Rune, both for transparency and for ease of uploading new versions of Outmatched.

oblador: Tried that, but build fails because the imports doesn't exist. Could perhaps rewrite it in a way where they are (not) resolved at runtime instead of build time.

oblador: Fixed in [1cf22d7](https://github.com/rune/rune-games-sdk/pull/149/commits/1cf22d72dd4e2adb4859a9686461487f54ef87b8)

helios1138: @oblador can you add the build/copy step similar to the one in Sudoku?

oblador: I would prefer if it worked the other way around, instead of this build having dependencies on private internal repos, we'd have a private internal repo that referenced this repo.

helios1138: Agreed to use a new repo for sounds, a separate deploy script in packages.json, a shared script file between games, temp checkout of repo via ssh link and fail for everyone

oblador: I've migrated to use this repo now 👍

helios1138 (approved):

PR #151, opened 2023-05-22
feat: `rune create` command

oblador:
Screenshot 2023-05-22 at 12 23 12


bfelbo (commented):
Awesome to add this 🥳

packages/rune-games-cli/src/flows/Create.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+function copyDir(srcDir: string, destDir: string) {
+  fs.mkdirSync(destDir, { recursive: true })
+
+  for (const file of fs.readdirSync(srcDir)) {
+    const srcFile = path.resolve(srcDir, file)
+    const destFile = path.resolve(destDir, file)
+    copy(srcFile, destFile)
+  }
+}
```
helios1138, line 48: I think you can just use `fs.cpSync()` with the `recursive` option

oblador: Docs marks it as `Experimental`

helios1138: sounds good

helios1138 (approved):

PR #152, opened 2023-05-22
feat: add examples demo

oblador:
Changes the TTT example to use new dev UI, adds a README and demos playable in the docs website.

### DIFF
--- a/docs/.gitignore
+++ b/docs/.gitignore
@@ -3,6 +3,7 @@
 
 # Production
 /build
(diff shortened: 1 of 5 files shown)


helios1138:
@oblador I get `Page Not Found` when I try to open the demo

oblador:
Yep, because this PR is adding it, so won't work until it's merged :D

helios1138 (approved):

PR #157, opened 2023-05-31
fix: use relative paths in vite template

oblador:


Gongreg (approved):
Lets add a comment explaining that this sets the path to relative

PR #158, opened 2023-05-31
docs: update readme with focus on multiplayer sdk

oblador:


bfelbo (approved):
LGTM!

Some suggestions:
1. Let's update the "Tiny SDK for making your HTML5 game run inside Rune" here and in the repo's about text to focus on multiplayer and what Rune provides more specifically?
2. Might be cool to call out the technologies used for each example like we do on https://developers.rune.ai/docs/multiplayer/examples to show the diversity of approaches we support?
3. Might be nice to add more direct links to the docs for non-API stuff, e.g. "Supported Games" and "Publishing your Game"?

oblador:
Since we have a separate documentation website, prefer if the content isn't too duplicated, and instead links to the other places where that has more info. So, I'm up for adding more links but not necessarily more content.

bfelbo:
Yes, that's what I meant :)

PR #164, opened 2023-05-31
docs: game time

Gongreg:


docs/docs/multiplayer/time.md
```diff
@@ -0,0 +1,82 @@
...
+  actions: {
+    requestGuess(_, {game}) {
+      game.guessRequestedAt = Rune.gameTimeInSeconds();
+    },
+    confirmGuess(_, {game}) {
+      game.guessDuration = Rune.gameTimeInSeconds() - game.guessDuration;
+    }
+  }
+})
+```
```
bfelbo: How about we use a quiz score example instead, where the score depends on the time it took to answer? I think that might be a more common use case :)

docs/docs/multiplayer/time.md
```diff
@@ -0,0 +1,82 @@
...
+## `timeSync` event
+
+Running timers and keeping them in sync requires quite a lot of complexity.
+Lets take an example:
+* Client executes an action when gameTime is at second 4.
+* Server receives the action, but server is already at second 5.
+* Every client receives that the action was executed at 5th second.
+
+In this scenario the client that executed the action will receive `timeSync` event, 
+indicating that something has changed due to action being executed at different time.
```
bfelbo: I'd suggest we reverse it to start by describing `timeSync` event, then give the example. We can also leave out the complexity part as Rune handles that and we wouldn't want anyone to think that they'd have to deal with the complexity :)

docs/docs/multiplayer/time.md
bfelbo: I think it might also be good to use examples that support multiple simultaneous players. We could consider using the examples from the initial time-syncing scope document?

docs/docs/multiplayer/time.md
```diff
@@ -0,0 +1,82 @@
+---
+sidebar_position: 61
+---
+
+# Time
```
bfelbo:
```suggestion
# Time-Based Logic
```

docs/docs/multiplayer/time.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Game time
```
bfelbo:
```suggestion
## Game Time
```

We capitalize subtitles

docs/docs/multiplayer/time.md
```diff
@@ -0,0 +1,82 @@
...
+
+If your game requires access to time, Rune exposes `Rune.gameTimeInSeconds()` that you can use inside logic.js.
+Calling it returns the amount of seconds that have passed since the start of the game.
+
+At the moment Rune allows timers with an accuracy of a second.
```
bfelbo:
```suggestion
Rune exposes `Rune.gameTimeInSeconds()` that you can use inside logic.js, which returns the amount of seconds that have passed since the start of the game. At the moment Rune provides time precision of a second, which should work well for most casual game purposes.
```

docs/docs/multiplayer/time.md
```diff
@@ -0,0 +1,82 @@
...
+    }
+  }
+})
+```
+
+## Update function
+
+If your game requires some logic to run every second, Rune allows you to provide `update` function inside `logic.js`.
+
+When game state is changed inside `update` function, `onChange` inside `client.js` is called with `update` event.
```
bfelbo:
```suggestion
You can provide an `update` function inside your `logic.js` file to run game logic every second. When game state is changed inside your `update` function, the `onChange` inside `client.js` is called with `update` event.

Here’s an example of how a game could implement that players have to make a move within 30 seconds or else their turn will pass:
```

bfelbo (commented):
Looks good! This is a tricky concept to explain well, added some suggestions!

docs/docs/multiplayer/time.md
Gongreg: I've switched to quiz example. I just have a feeling that it is a bit too complicated to easily indicate the usage of gameTimeInSeconds.

docs/docs/multiplayer/time-based-logic.md
```diff
@@ -0,0 +1,104 @@
...
+      game.clickedAt = Rune.gameTimeInSeconds();
+    },
+  },
+})
+```
+
+* PlayerA calls action click when game time is at second 4. PlayerA receives `onChange` call with `click` action as an argument. PlayerA `game.clickedAt = 4`.
+* Server receives the action, but server is already at second 5. Server processes the action.
+* Every client receives that the action was executed at 5th second. 
+PlayerA `onChange` is called with `timeSync` event, playerB `onChange` is called with `click` action. Both players `game.clickedAt = 5`.
```
bfelbo:
```suggestion
* PlayerA calls action click when game time is at second 4. PlayerA receives `onChange` call with `click` action as an argument, making PlayerA see `game.clickedAt = 4`.
* Server receives and processes the action at game time second 5.
* Every client receives that the action was executed at 5th second. 
PlayerA's `onChange` function is called with `timeSync` event to reconcile that the server processed the action at second 5 (and the server holds the truth). In this way, both players end up with `game.clickedAt = 5`.
```

bfelbo (approved):
Looks great! 🥳

PR #165, opened 2023-05-31
build: use https for copy-assets

bfelbo:
Minor internal change for deploying example games.


helios1138 (changes requested):
What's the motivation here? 🙂 SSH is the recommended way to access git.

helios1138 (approved):
as discussed, will merge just the comment

PR #167, opened 2023-06-14
refactor: multiplayer focus

bfelbo:
Change multiplayer examples to be top-level and remove singleplayer code as it's being deprecated (see warning on https://developers.rune.ai/docs/singleplayer/intro).


README.md
```diff
@@ -24,9 +24,9 @@ SDK for building multiplayer HTML5 games that run inside Rune on [Android](https
...
+| [Demo](https://developers.rune.ai/multiplayer/examples/tic-tac-toe/), [Source](https://github.com/rune/rune-games-sdk/tree/staging/examples/tic-tac-toe) | [Demo](https://developers.rune.ai/multiplayer/examples/outmatched/), [Source](https://github.com/rune/rune-games-sdk/tree/staging/examples/outmatched) | [Demo](https://developers.rune.ai/multiplayer/examples/tic-tac-toe/), [Source](https://github.com/rune/rune-games-sdk/tree/staging/examples/sudoku) | [Demo](https://developers.rune.ai/multiplayer/examples/pinpoint/), [Source](https://github.com/rune/rune-games-sdk/tree/staging/examples/pinpoint) |
```
Gongreg: Is the path here correct to the demos?

bfelbo: Yes, until we change the docs to no longer have singleplayer docs. We could consider also doing that now I guess 🤠 But I'd say let's do it separately as this is a big PR already and there's some choices to be made there.

bfelbo: Ah, nvm. You're right 👍

bfelbo: Fixed in https://github.com/rune/rune-games-sdk/pull/167/commits/277961757cb0cd1f3186c583eaca938cb817360b.

Gongreg (approved):

PR #168, opened 2023-06-14
docs: make outmatched image smaller

Gongreg:


bfelbo:
We'll try to fix this another way as discussed :)

PR #179, opened 2023-07-12
feat: request twitter handle when registering

Gongreg:
This will show a step to enter twitter handle after requesting display name.

Leaving empty handle is treated as skipping it.
Also in case user exits the cli while requesting for twitter handle, we'll also simply treat it as a skip.


helios1138 (changes requested):
not sure I understand the point of `isRegistering` since you only set it to true when handle is empty, so it's the same condition essentially

and I think we don't need to call updateDevTeamById twice.

just show login when there is no handle, ask for both handle and twitter, call updateDevTeamById once but don't provide twitter field if it's empty.

helios1138 (approved):

PR #185, opened 2023-07-21
feat: ask if ready for release during rune upload

mieszko4:
During `rune upload` ask if game is ready for release so that dev can send draft versions of the game.
This is useful for devs wanting to share playtesting links while developing the game.
Once dev decides that game version is ready for release, the reviewing of the game can begin.

### Test plan
- [x] Upload version which is not ready for release
- [x] Upload version which is ready for release


bfelbo (approved):
Looks great! 🥳

packages/rune-games-cli/src/flows/Upload/ReadyForReleaseStep.tsx
```diff
@@ -0,0 +1,41 @@
...
+  useEffect(() => {
+    if (readyForRelease !== undefined) onComplete(readyForRelease)
+  }, [readyForRelease, onComplete])
+
+  const label = useMemo(() => {
+    if (readyForRelease === undefined) return "Is your game ready for release?"
+
+    return readyForRelease
+      ? "Game is ready for release"
+      : "Game is not ready for release"
```
bfelbo:
```suggestion
    return readyForRelease
      ? "Yes, please review and release it"
      : "No, just uploading for playtesting"
```

mieszko4: Sounds good. Just note that the part you suggest to change is what will be displayed **after** dev already made the choice.

During the choice we still display `Is your game ready for release?` with `No`/`Yes`.

EDIT: We cannot make this change because after the question is answered we do not show the question anymore (this is how all questions are implemented at the moment).

So it would make sense to be more like:
`Game is ready, please review and release it` / `Game is not ready, just uploading for playtesting`.

bfelbo: I see, thanks. We could use the text you suggested here and the text I suggested for the “options” variable below.

mieszko4: After discussion with @bfelbo, having so long answers and my previous suggestion may be confusing for devs hence changing only info after the question is answered (https://github.com/rune/rune-games-sdk/pull/185/commits/4d633acf5ec8770c43ec488e7507b752057c5b89)

PR #190, opened 2023-07-24
fix: eslint typescript support + recommended config change

Gongreg:
* Fixes an issue where typescript files were not checked for restrictedGlobals.
* Removes check for Map/Set.
* Updated tests to test code as typescript/javascript + module/script.
* Changed recommended eslint config, now it always treats any logic.js/logic.ts file as module. If users want to do additional customization we should always remember to link to https://developers.rune.ai/docs/advanced/server-side-logic/
* Updated game template to use the rune eslint plugin.

### DIFF
--- a/packages/eslint-plugin-rune/src/index.ts
+++ b/packages/eslint-plugin-rune/src/index.ts
@@ -129,7 +129,7 @@ const logicModuleConfig: ESLint.ConfigData = {
 export const configs: ESLint.Plugin["configs"] = {
   recommended: {
     globals: {
-      Rune: true,
+      Rune: "readonly",
     },
     overrides: [
       {
--- a/packages/eslint-plugin-rune/src/index.ts
+++ b/packages/eslint-plugin-rune/src/index.ts
@@ -79,42 +79,6 @@ const logicConfig: ESLint.ConfigData = {
     ecmaVersion: 2021,
   },
   globals: {
-    // See https://github.com/eslint/eslint/tree/main/conf/globals.js
-    // es3
-    Array: "readonly",
-    Boolean: "readonly",
-    Error: "readonly",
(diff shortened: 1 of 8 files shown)


packages/eslint-plugin-rune/src/index.ts
```diff
@@ -23,34 +23,67 @@ const restrictedSyntaxBase = [
...
+  "Performance",
+  "Intl",
+]
+
 const logicConfig: ESLint.ConfigData = {
   plugins: ["rune"],
   parserOptions: {
     ecmaVersion: 2021,
   },
   globals: {
```
bfelbo, line 81: Would be great with a comment explaining why there's both this and `restrictedGlobals` at the top and why these are not part of `restrictedGlobals` :)

packages/eslint-plugin-rune/test/config/parent-scope-mutation.spec.js
```diff
@@ -1,10 +1,8 @@
 const { createConfigTester } = require("../createConfigTester")
 
-const test = createConfigTester({
-  extends: "plugin:rune/logic",
-})
+const test = createConfigTester()
 
-test("variable scope", {
+test(`variable scope`, ({ type }) => ({
```
bfelbo:
```suggestion
test("variable scope", ({ type }) => ({
```

bfelbo (approved):
Looks great! 💯

packages/eslint-plugin-rune/src/index.ts
Gongreg: I've ran the tests, and it seems like these were completely redundant (covered by parentScope & noGlobalAssign). Removed it altogether.

bfelbo: Sounds great 👍

PR #194, opened 2023-07-26
fix: allow globalThis + update template to run linter when building

Gongreg:
* Allows to use globalThis.
* Updates vite template to run lint when building.


bfelbo (approved):
Great improvements!

PR #195, opened 2023-07-26
fix: allow set

Gongreg:
Set was not allowed due to env missing.


bfelbo (approved):
Great to add support for this 👍

PR #209, opened 2023-08-08
feat: upload preview img by admin using rune update-info

mieszko4:
Ask for optional preview img when admin runs `rune update-info`.

### Test plan
- [x] Run `rune update-info` as admin and use default data.
- [x] Run `rune update-info` as admin and update all data. Make sure new logo and preview img got uploaded
- [x] Rune `rune update-info` as normal user and use default data.
- [x] Rune `rune update-info` as normal user and update all data. Make sure you cannot see preview img and that all other data got successfully updated.


helios1138 (approved):
💯

PR #210, opened 2023-08-10
fix: update both title and description when running rune update-info

mieszko4:
Currently, when both title and description are being updated during `rune update-info`, only `title` gets updated.
This PR fixes this bug.

### Test plan
- [x] Run `rune update-info` and select game. Make sure both title and description are pre-filled
- [x] Update both title and description and ignore logo. Make sure both title and description are saved

### DIFF
--- a/packages/rune-games-cli/src/flows/Update/UpdateGameStep.tsx
+++ b/packages/rune-games-cli/src/flows/Update/UpdateGameStep.tsx
@@ -63,11 +63,15 @@ export function UpdateGameStep({ gameId }: { gameId: number }) {
   }, [updateGameError])
 
   useEffect(() => {
+    // Set initial values of game
     if (game) {
-      setTitle(game.title)
-      if (game.description) setDescription(game.description)
+      if (!titleSubmitted) setTitle(game.title)
+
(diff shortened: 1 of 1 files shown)


helios1138 (commented):
It might be more explicit/clear if we do it like this?
```ts
const [initialValuesSet, setInitialValuesSet] = useState(false)

useEffect(() => {
  if (game && !initialValuesSet) {
    setTitle(game.title)
    if (game.description) setDescription(game.description)
    setInitialValuesSet(true)
  }
}, [game, initialValuesSet])
```

mieszko4:
> It might be more explicit/clear if we do it like this?
> 
> ```ts
> const [initialValuesSet, setInitialValuesSet] = useState(false)
> 
> useEffect(() => {
>   if (game && !initialValuesSet) {
>     setTitle(game.title)
>     if (game.description) setDescription(game.description)
>     setInitialValuesSet(true)
>   }
> }, [game, initialValuesSet])
> ```

Fixed in https://github.com/rune/rune-games-sdk/pull/210/commits/381ec61dc93b52b830267b269fa51552dd802d16

helios1138 (approved):

PR #239, opened 2023-10-06
feat: show my role and other game devs in game item

mieszko4:
Restyle game item to show my role and all game devs.


mzbyszynski (approved):
looks good to me.

PR #242, opened 2023-10-10
refactor: add eslint rules in every example game

Gongreg:


examples/oink-oink/.eslintrc.cjs
```diff
@@ -0,0 +1,16 @@
+module.exports = {
```
helios1138, line 1: I would use JSON or YAML for the config since we don't need any dynamic stuff here, but FFTI

helios1138 (approved):

PR #243, opened 2023-10-16
fix: replace parent scope with global scope

Gongreg:
This should allow users to use things like .map, .forEach, and then modify `game` or any nearby created object inside of it.


packages/eslint-plugin-rune/test/config/global-scope-mutation.spec.js
```diff
@@ -87,9 +110,22 @@ test("variable scope", ({ type }) => ({
...
+    `
+    const arr = [1, 2, 3];
+    
+     function globalReference() {
+       return {
+         arr
+       };
+     }
+    
+    function abc() {
```
bfelbo:
```suggestion
    const arr = [1, 2, 3];
    
    function abc() {
```

bfelbo (approved):
Awesome to fix this!

PR #252, opened 2023-10-20
feat: Cube Rush improvements

mieszko4:
Cube Rush
* Add controls tap indicator
* Add keyboard controls
* Add transition on stats
* Refactor to use classList instead of style
* Optimize collision detection and size of game state


bfelbo (approved):
💯

PR #262, opened 2023-10-26
feat: dont minify logic file

Gongreg:
Currently we run `logic.js` validation before uploading the game.

There is a big chance that user didn't make any mistakes in their logic code, but they've imported some 3rd party code, which is written not following the necessary rules.

Since we minify the logic.js code, its becomes more or less impossible to understand what is going on.

This PR stops minifying logic.js code.


bfelbo (approved):
Great to avoid minifying logic.js! 🥳 Don't know that much about vite internals and plugins, but I trust that you've tested this well.

PR #263, opened 2023-10-27
feat: optimize neon snake

bfelbo:
Fewer cells in collision grid.


helios1138 (commented):
@bfelbo I don't think our logic can support the size >= 2x move

helios1138:
ah, no ignore that, the limit is actually <= 0.5x

bfelbo:
Cool. Everything also worked great when testing :)

PR #278, opened 2023-12-07
fix: cuberush improvements

Gongreg:
* Added car blinking when colliding with a cube.
* Using instanced mesh for cubes. This way only 1 GPU call is done for them.
* Pregenerated maps to reduce the game state size.
* Removed background since it is already defined on body and was leading to flickering on desktop chrome.
* Removed touch-event: none from body, since it was leading to performance issues on chrome android: https://stackoverflow.com/a/35668492


bfelbo (approved):
Great improvements!

PR #279, opened 2023-12-11
fix: Neon snake improvements

Gongreg:
* End game after 5 points
* Increase turnDegreesPerTick by 50% (close enough)
* Switched collision grid from `{ [index: number]: boolean }` to `Record<Point["x"], Record<Point["y"], boolean>>`. This is a significant performance increase when using `mutatitve`.
* Support tap to turn in dev UI
* Fix self-colliding. Should completely prevent self collisions now by keeping last 3 visited collision points. This way if snake visits a point it just was at, we'll allow it. During normal gameplay its impossible to turn so quickly, so we can use this as a solution.


examples/neon-snake/src/components/BoardScreen/drawBoard.ts
```diff
@@ -5,14 +5,61 @@ import { drawDeadEnd } from "./drawDeadEnd.ts"
 import { drawAvatar } from "./drawAvatar.ts"
 import { getOptimisticStartSection } from "./getOptimisticStartSection.ts"
 
+////Draws a board showing collision grid. Useful for testing
```
bfelbo: FFTI: Let's keep the function as-is and just comment out the use of it?

examples/neon-snake/src/logic/logicConfig.ts
```diff
@@ -1,10 +1,12 @@
...
 export const boardSize = { width: 600, height: 900 }
 export const movePixelsPerTick = 2.5
-export const turnDegreesPerTick = 3
+export const turnDegreesPerTick = 5
 export const gapChancePerTick = 0.01
 export const gapLengthTicks = 20
 export const minTicksToNextGap = 30
 export const colors = ["#BCFE00", "#10D4FF", "#FF32D2", "#FF9C27"]
+
+export const ALLOWED_COLLISION_POINTS = 3
```
bfelbo:
```suggestion
export const allowedCollisionPoints = 3
```

FFTI: To keep in same style as remaining code

bfelbo (approved):
Great improvements 💯

examples/neon-snake/src/components/BoardScreen/drawBoard.ts
Gongreg: Typescript throws an error:
`TS6133  is declared but its value is never read` due to `"noUnusedLocals": true,` rule being enabled in this example game. I'll just keep the function commented out.

examples/neon-snake/src/logic/updatePlaying/checkLatestSectionForCollisions.ts
```diff
@@ -1,7 +1,46 @@
...
+) {
+  //Initialize collision grid if it does not exist yet
+
+  if (!collisionGrid[collisionPoint.x]) {
+    collisionGrid[collisionPoint.x] = {}
+  }
+
+  collisionGrid[collisionPoint.x][collisionPoint.y] = true
+
+  if (!isCollisionPointRecentlyVisitedBySnake(snake, collisionPoint)) {
```
helios1138, line 24: I think you probably don't need to check for this again here

Gongreg: This is a check to avoid pushing in same value multiple times. I'll add a comment

helios1138: Then can we reuse `isSameCollisionPointAsBefore`? I'm just worried about calling an array lookup (`.some()`) more than necessary.

Gongreg: I'm not sure what you mean. Each `markCollisionGrid` call is done with different `collisionPoint`.

helios1138: > Each markCollisionGrid call is done with different collisionPoint.

Then why do you need to "check to avoid pushing in same value multiple times."?

Gongreg: Because the snake can be in the same collision point multiple `update` calls in a row.

PR #281, opened 2023-12-12
refactor: rewrite neon snake on react-pixi

Gongreg:
Using pixi.js seems to lead to significant performance improvements, especially on mobile devices.

There is one drawback, using Graphics I'm unable to implement glow effect that would be efficient and would look okay.


bfelbo (approved):
Great to make the game more performant! 💯

PR #285, opened 2023-12-15
Tic-Tac-Toe: disable cells when it's not your turn

helios1138:


bfelbo (approved):
Sounds good! I've tweaked it a little bit.

PR #292, opened 2024-01-02
feat: interactive game list with details

mieszko4:
* Sort games alphabetically with your games on top
* Make `rune list` interactive
* Add game details screen


packages/rune-games-cli/src/flows/List/Details.tsx
```diff
@@ -0,0 +1,37 @@
+import { Text, Box } from "ink"
+import React from "react"
+
+import { Step } from "../../components/Step.js"
+import { useGame } from "../../gql/useGame.js"
+import { useMe } from "../../gql/useMe.js"
+
+export function Details({ gameId }: { gameId: number }) {
+  const { me } = useMe()
+  const { game, gameLoading } = useGame(gameId)
```
mzbyszynski: Maybe `me`, `game` and `gameLoading` should be passed in as props, since they are already fetched by the parent `List` component?

packages/rune-games-cli/src/flows/List/Details.tsx
```diff
@@ -0,0 +1,37 @@
...
+  const gameDevs = game?.gameDevs.nodes
+  const gameDevMe = gameDevs?.find((gameDev) => gameDev.userId === me?.devId)
+
+  if (gameLoading) {
+    return <Step status="waiting" label="Loading game" />
```
mzbyszynski: Initially it seemed strange to have a waiting status in `Details` and also in `List`, but I see that the labels are different. Unless the intent is that this `Details` component will be reused somewhere else outside of `List`, it might be nice to centralize all the "loading" status displays in one place (`List`) instead of two.

mzbyszynski (approved):
Looks reasonable to me. I have a few minor suggestions, but nothing major. Might be nice to have some tests for some of the busier parts of the code, like `useGames`, but I see that this package doesn't have many tests so it might be more trouble than it is worth getting them set up properly.

packages/rune-games-cli/src/flows/List/Details.tsx
mieszko4: I was considering this but then I concluded that it would be better to adjust our apollo fragments and cache policy so that `useGame` and `useMe` instantly return cached data. But sure, let's do that right away until we figure out that out.

packages/rune-games-cli/src/flows/List/Details.tsx
mieszko4: Done in https://github.com/rune/rune/pull/292/commits/ff22bccf453cd1423dad25e998e6b233f53f0e68

packages/rune-games-cli/src/flows/List/Details.tsx
mieszko4: Done in https://github.com/rune/rune/pull/292/commits/ff22bccf453cd1423dad25e998e6b233f53f0e68

PR #294, opened 2024-01-03
refactor: useGames

mieszko4:
* Move out `gameItemLabel` and `useMyGames` into libs
* Add tests for `gameItemLabel` and `useMyGames`


mzbyszynski (approved):
Thanks for adding those tests. Great work!

PR #296, opened 2024-01-05
feat(eslint): no setTimeout

bfelbo:
Realized that our eslint plugin didn't catch use of `setTimeout` so I added it as it's not allowed since Rune logic must be synchronous.


Gongreg (approved):

bfelbo:
This needs some changes, closing it for now

PR #299, opened 2024-01-15
docs: rune update-members

mieszko4:
Add docs for Collaboration and `rune update members`

Depends on https://github.com/rune/rune/pull/298


bfelbo (approved):
Great to explain more how this all works 💯

PR #306, opened 2024-01-18
fix: use directory name if relative path

chrisrabe:
Added a fix where if they create a rune app with a relative path in their name (e.g. `.` or `..`), then it would update the package name to be the directory name associated to the relative path.

### DIFF
--- a/packages/rune-games-cli/src/flows/Create.tsx
+++ b/packages/rune-games-cli/src/flows/Create.tsx
@@ -1,7 +1,6 @@
 import fs from "fs"
 import { Box, Text } from "ink"
 import { UncontrolledTextInput } from "ink-text-input"
-import { kebabCase } from "lodash"
 import path from "path"
 import React, { useCallback, useEffect } from "react"
 import { fileURLToPath } from "url"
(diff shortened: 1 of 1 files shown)


chrisrabe:

packages/rune-games-cli/src/flows/Create.tsx
```diff
@@ -65,6 +65,21 @@ function emptyDir(dir: string) {
   }
 }
 
+const isRelativePath = (path: string) =>
+  path.startsWith(".") || path.startsWith("..")
+
+const getPackageName = (targetDir: string) => {
+  if (!isRelativePath(targetDir)) return targetDir
```
Gongreg: Should we always use this function even if the directory is not relative?
Directory name could contain spaces or other symbols that are not allowed in package name.

chrisrabe: Yeah I guess we can add a ternary operator where pkg.name is assigned. I thought encapsulating all the package name logic into one function is more readable.

I accounted for directory name containing spaces or other symbols by replacing non-hyphen characters with hyphens. I didn't account for when the user types in weird characters for targetDir though

Gongreg: Sorry, I meant specifically removing the line 72, so that your logic is always used. I guess then we need to modify line 74 to support both relative and non relative path. That way we'll always make sure to use a valid name for package.json

chrisrabe: Ah yep. I updated it so that it transforms it to valid package name and only performs directory operations if it's a relative path

Gongreg (approved):
I've updated the pr to always use absolute path.

Gongreg:
Thanks for the contribution @chrisrabe !

PR #307, opened 2024-01-18
feat: add human err when updating admin of the game

mieszko4:
Add human err when updating last active admin of the game using `rune update-members`


bfelbo (approved):
Great with better errors 👍

PR #308, opened 2024-01-22
docs: add autogenerated hardcoded hashtags

helios1138:
This will make it easier to keep hashtag links alive in case of wording changes in the section headers


bfelbo (approved):
Sounds good 👍

PR #316, opened 2024-02-14
feat: Add more human err msgs to rune update-members

mieszko4:
Add human err msg in case invitee has no email or invitee's email is blocked.


bfelbo (approved):
Great to have more intuitive error msgs 👍

PR #319, opened 2024-02-18
feat: allow minification of logic.js via env variable RUNE_MINIFY_LOGIC

dmecke:
Allows using the env variable RUNE_MINIFY_LOGIC to minify `logic.js`. By default it will not be minified, setting the variable to `1` will minify the file.

### DIFF
--- a/packages/vite-plugin-rune/src/terser.ts
+++ b/packages/vite-plugin-rune/src/terser.ts
@@ -48,7 +48,7 @@ export function terserPlugin(): Plugin {
         Object.keys(bundle).map(async (name) => {
           if (
             !shouldMinify ||
-            (name === "logic.js" && !process.env.RUNE_MINIFY_LOGIC)
+            (name === "logic.js" && process.env.RUNE_MINIFY_LOGIC !== "1")
           ) {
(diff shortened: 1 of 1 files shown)


dmecke:

packages/vite-plugin-rune/src/terser.ts
```diff
@@ -46,7 +46,10 @@ export function terserPlugin(): Plugin {
 
       await Promise.all(
         Object.keys(bundle).map(async (name) => {
-          if (!shouldMinify || name === "logic.js") {
+          if (
+            !shouldMinify ||
+            (name === "logic.js" && !process.env.RUNE_MINIFY_LOGIC)
```
bfelbo:
```suggestion
            (name === "logic.js" && process.env.RUNE_MINIFY_LOGIC !== "1")
```

Would be great to make `RUNE_MINIFY_LOGIC = 1` still minify the logic.js file :)

**EDIT:** Fixed it to 1 instead of 0

bfelbo (approved):
One minor suggestion, otherwise looks great!

PR #320, opened 2024-02-19
feat: optional param to enable logic minification

Gongreg:
Adds an optional param to enable logic minification.

Replaces the https://github.com/rune/rune-multiplayer-web-games/pull/319 PR.

### DIFF
--- a/packages/rune-games-cli/template-vite-react-ts/vite.config.ts
+++ b/packages/rune-games-cli/template-vite-react-ts/vite.config.ts
@@ -10,6 +10,9 @@ export default defineConfig({
   plugins: [
     qrcode(), // only applies in dev mode
     react(),
-    rune({ logicPath: path.resolve("./src/logic.ts") }),
+    rune({
+      logicPath: path.resolve("./src/logic.ts"),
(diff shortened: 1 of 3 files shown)


packages/rune-games-cli/template-vite-react-ts/vite.config.ts
```diff
@@ -10,6 +10,9 @@ export default defineConfig({
   plugins: [
     qrcode(), // only applies in dev mode
     react(),
-    rune({ logicPath: path.resolve("./src/logic.ts") }),
+    rune({
+      logicPath: path.resolve("./src/logic.ts"),
+      enableLogicMinification: false, // If your logic reaches the allowed limit, you can enable this flag to minify it. This will make it significantly more difficult to detect validation issues.
```
bfelbo:
```suggestion
      minifyLogic: false, // This flag can be used if your logic reaches the allowed limit. However, it will make it significantly more difficult to detect validation issues.
```

bfelbo (approved):
Good idea to do it as a vite plugin variable.

I'd suggest we avoid shorten/simplify the key

PR #329, opened 2024-03-15
feat: better create command

Gongreg:
Added a step that suggests to install dependencies using either npm or alternative (depending on process.env.npm_config_user_agent).

### DIFF
--- a/packages/rune-games-cli/src/flows/Create.tsx
+++ b/packages/rune-games-cli/src/flows/Create.tsx
@@ -73,12 +73,12 @@ function emptyDir(dir: string) {
 const getPackageName = (targetDir: string) => {
   const fullTargetDirPath = path.resolve(targetDir)
 
-  const lastFolderName = fullTargetDirPath.split(path.sep).pop()
+  const lastDirectoryName = fullTargetDirPath.split(path.sep).pop()
 
-  if (!lastFolderName) return "rune-game-template" // in-case they put in a root relative path
+  if (!lastDirectoryName) return "rune-game-template" // in-case they put in a root relative path
 
   // Replace any non-hyphen characters (like spaces or underscores) with hyphens
-  return lastFolderName.replace(/[^a-zA-Z0-9]/g, "-")
+  return lastDirectoryName.replace(/[^a-zA-Z0-9]/g, "-")
 }
 
 export function Create({ args }: { args: string[] }) {
@@ -176,7 +176,7 @@ export function Create({ args }: { args: string[] }) {
         label={
           step > Steps.Target
             ? `Will create a game in "${targetDir}"`
-            : "Game folder name"
+            : "Game directory name"
         }
         view={
           step <= Steps.Target && (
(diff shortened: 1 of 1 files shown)


packages/rune-games-cli/src/flows/Create.tsx
```diff
@@ -133,16 +159,24 @@ export function Create({ args }: { args: string[] }) {
...
   return (
     <Box flexDirection="column">
       <Step
         status={step > Steps.Target ? "success" : "userInput"}
         label={
           step > Steps.Target
-            ? `Will create a project in "${targetDir}"`
-            : "Project name"
+            ? `Will create a game in "${targetDir}"`
+            : "Game folder name"
```
kevglass: Would it be better to stick to "directory" here - folder is quite non-dev language.

packages/rune-games-cli/src/flows/Create.tsx
```diff
@@ -153,58 +187,106 @@ export function Create({ args }: { args: string[] }) {
...
+
+      {step >= Steps.Creating && (
         <Step
-          status={step > Steps.Overwrite ? "success" : "userInput"}
+          status={step > Steps.Creating ? "success" : "waiting"}
           label={
             step > Steps.Overwrite
-              ? `Will overwrite existing folder`
-              : `Target directory "${targetDir}" is not empty. Remove existing files and continue?`
+              ? `File structure created!`
```
kevglass: I think we're creating a directory/folder structure, rather the structure of a file?

packages/rune-games-cli/src/flows/Create.tsx
```diff
@@ -153,58 +187,106 @@ export function Create({ args }: { args: string[] }) {
...
+              ? `Dependencies installed successfully!`
+              : step === Steps.InstallError
+              ? `Failed to install dependencies`
+              : `Installing dependencies...`
+          }
+        />
       )}
-      {step === Steps.Done && (
+
+      {(step === Steps.SkipInstall || step === Steps.InstallError) && (
```
kevglass, line 276: Are we telling them they can start the project even after an install error?

packages/rune-games-cli/src/flows/Create.tsx
Gongreg: We've been going with @bfelbo on this one for a bit. I'm okay with both :)

bfelbo: Don't have a preference either. Sounds good to do "Game directory" 👍

kevglass: The preference for me comes from thinking of devs as command line users where as folders are for UI users. That said, I'm probably just dated :)

Gongreg: renamed to directory.

kevglass (approved):

packages/rune-games-cli/src/flows/Create.tsx
Gongreg: Yes, in case something goes wrong they can try to do npm/yarn install manually. We do say to do the install first.

packages/rune-games-cli/src/flows/Create.tsx
Gongreg: changed the content, thanks.

PR #334, opened 2024-03-20
feat: detect external dependencies

Gongreg:
Detects external dependencies used in any file imported by logic.

Works both in serve and build commands.

In serve command it works as a watcher, any change done (removing/adding a dependency) should be reflected in the terminal.

In build command it also lists the dependencies to terminal, but also lists them in a comment on top of logic file.

Added a global whitelist that users can contribute to in case they use a library that works correctly with game logic.

Added a option param `ignoredDependencies` in the plugin, in case user doesn't want to contribute to the repo, but still wants the warning message be gone.


kevglass (commented):
LGTM - gone through the code - crazy but awesome :)

One question, will this cope with imports of resources? I've seen a few bits of source that import JSON level files etc.

Gongreg:
Great question. I've added a test to verify that it works with json.

kevglass (approved):

PR #348, opened 2024-04-08
Add best practice for control state

kevglass:


docs/docs/publishing/best-practices.md
```diff
@@ -26,6 +26,10 @@ Players expect any level information to persist indefinitely, but iOS/Android ma
 
 Rune shows an animation while your game is loading. It's a smoother player experience to only have one progress bar to load your game. Avoid having a loading screen in your game by waiting with calling `Rune.initClient()` until your game has fully completed loading.
 
+## Send control state rather than positions and only when they change {#send-controls-on-change}
+
+To minimize the amount of data on the network send the control state, e.g. whether the user is pressing left or right, in the actions. Only send an action when the control changes to avoid hitting the limit on actions per second.
+
```
bfelbo:
```suggestion
## Send player input rather than player state in actions {#send-player-input}

We recommend sending player input in actions (e.g. "turning left") rather than player state (e.g. the player's position). This minimizes network data and allows other clients to simulate ahead, thereby making your game work better in bad network settings. Similarly, only send an action when the input changes instead of sending the same player input repeatedly.

```

bfelbo (approved):
Great to clarify this!

I've made a suggestion to use "player input" instead of "control state" as I think it'll be easier to understand, especially for ESL devs. Curious to get your thoughts :)

kevglass:
> Great to clarify this!
> 
> I've made a suggestion to use "player input" instead of "control state" as I think it'll be easier to understand, especially for ESL devs. Curious to get your thoughts :)

It's really interesting/difficult trying to understand what will make sense to ESL readers. The language isn't "simple" it's more common word usage I think.

PR #357, opened 2024-04-18
fix: allow relative import

Gongreg:
Fixes an issue where '../' import was not correctly handled.


packages/vite-plugin-rune/test/fixtures/outer.ts
```diff
@@ -0,0 +1,3 @@
+export function outer() {
```
kevglass, line 1: FFTI: Not sure inner and outer are the best names, but can't think of anything better!

kevglass (approved):
LGTM

PR #359, opened 2024-04-29
feat: use css grid for board rendering in tic-tac-toe

helios1138:


bfelbo (approved):
LGTM 👍 

Let's also update `packages/dusk-cli/templates` (we keep these in sync during transition phase).

PR #369, opened 2024-05-02
fix: logic client html

Gongreg:
Updates generated html file to contain following format:

```
<script type="module" crossorigin src="./logic.js"></script>
<script type="module" crossorigin src="./client.js"></script>
```


packages/vite-plugin-rune/src/index.ts
```diff
@@ -30,9 +34,19 @@ export default function runePlugin(options: ViteRunePluginOptions): Plugin[] {
     throw new Error(`Rune logic file "${options.logicPath}" not found`)
   }
 
+  let runePkgPath: string
+  try {
+    runePkgPath = require.resolve("rune-games-sdk/package.json")
```
kevglass, line 39: Should this be coping with dusk-games-sdk ?

kevglass (approved):

PR #375, opened 2024-05-05
feat: React template no extra style import

bfelbo:
Just removes an unnecessary style import in the React template.


Gongreg (approved):
Ah, the import was duplicated. Thanks

PR #382, opened 2024-05-08
Add landscape flag to game metadata extracted during upload

helios1138:

### DIFF
--- a/packages/dusk-cli/src/lib/extractMultiplayerMetadata.ts
+++ b/packages/dusk-cli/src/lib/extractMultiplayerMetadata.ts
@@ -6,6 +6,7 @@ const regexes = {
   updatesPerSecondDefined: /updatesPerSecond\s*:/,
   updatesPerSecond: /updatesPerSecond\s*:\s*?([0-9]+)/,
   inputDelay: /inputDelay\s*:\s*?([0-9]+)/,
+  landscape: /landscape\s*:\s*?(true|false)/,
 }
 
 export function extractMultiplayerMetadata(logicJsContent: string) {
@@ -14,6 +15,9 @@ export function extractMultiplayerMetadata(logicJsContent: string) {
 
   const maxPlayersString = logicJsContent.match(regexes.maxPlayers)?.at(1)
   const maxPlayers = maxPlayersString ? parseInt(maxPlayersString) : undefined
+
+  const landscapeString = logicJsContent.match(regexes.landscape)?.at(1)
+  const landscape = landscapeString ? landscapeString === "true" : undefined
 
   const updatesPerSecondDefined =
     logicJsContent.match(regexes.updatesPerSecondDefined) !== null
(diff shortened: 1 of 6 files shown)


packages/dusk-cli/src/lib/extractMultiplayerMetadata.ts
```diff
@@ -15,6 +16,9 @@ export function extractMultiplayerMetadata(logicJsContent: string) {
   const maxPlayersString = logicJsContent.match(regexes.maxPlayers)?.at(1)
   const maxPlayers = maxPlayersString ? parseInt(maxPlayersString) : undefined
 
+  const landscapeString = logicJsContent.match(regexes.landscape)?.at(1)
+  const landscape = landscapeString ? landscapeString === "true" : undefined
```
mzbyszynski, line 20: Couldn't this just be `const landscape = landscapeString === "true"`? Or is it important that it be `undefined` and not `false`?

helios1138: It's just for consistency. You can see that all other fields are defined as optional in the resulting type. We could make them required, but we should make all of them required in that case 🙂

If I had to assign a (weak) meaning to this (retroactively), it would be this: since it's a parser, it outputs what is literally defined by the dev in the code, dev can set it to false or he can omit it. If it ever becomes important to know which choice did the dev make, it's better to not erase the information at the source. We can always cast to required type down the line (in fact we do that when writing to the DB).

mzbyszynski: makes sense to me!

mzbyszynski (approved):
looks good to me, for what is is worth. I have no idea what this code does though, so would be good to get @Gongreg's blessing too

packages/dusk-cli/src/lib/extractMultiplayerMetadata.ts
```diff
@@ -6,6 +6,7 @@ const regexes = {
   updatesPerSecondDefined: /updatesPerSecond\s*:/,
   updatesPerSecond: /updatesPerSecond\s*:\s*?([0-9]+)/,
   inputDelay: /inputDelay\s*:\s*?([0-9]+)/,
+  landscape: /landscape\s*:\s*?(true|false)/,
```
kevglass, line 9: @Gongreg are we going to see the same issue with !0 after minification here that was on the persistence property? (I think you turned it off for Pinpoint?)

kevglass (approved):

packages/dusk-cli/src/lib/extractMultiplayerMetadata.ts
Gongreg: We are going to see the same issue.

Gongreg: @helios1138 , we can simplify it to:
```/landscape\s*:\s*true/```
Because only care if the flag is set to true.

helios1138: We could, but I keep the approach [for consistency](https://github.com/rune/rune-multiplayer-web-games/pull/382#discussion_r1594712728) unless we want to change all the other fields as well now.

PR #389, opened 2024-05-17
fix: windows input fix

Gongreg:
Fixes an issue where users were not able to ender game name


bfelbo (approved):
I'm not that familiar with ink, but seems good to me

PR #399, opened 2024-06-13
feat: validate game config without regex

Gongreg:


packages/dusk-cli/src/lib/validateGameFiles.ts
```diff
@@ -257,85 +281,64 @@ async function validateMultiplayerLogicJsContent({
...
-        multiplayerMetadata.inputDelay > 50
-      ) {
-        errors.push({
-          message: "logic.js: inputDelay must be undefined or between 0 and 50",
-        })
-      }
+    if (gameConfig.updatesPerSecond === undefined) {
+      errors.push({
+        message:
+          "logic.js: updatesPerSecond must be a constant (updatesPerSecond: 1-30)",
```
kevglass, line 332: Is this true now?

packages/rune-games-cli/src/lib/validateGameFiles.ts
```diff
@@ -243,85 +281,64 @@ async function validateMultiplayerLogicJsContent({
...
-        multiplayerMetadata.updatesPerSecond > 30
-      ) {
-        errors.push({
-          message:
-            "logic.js: updatesPerSecond must be undefined or between 1 and 30",
-        })
-      }
-    }
-
-    if (multiplayerMetadata.inputDelay !== undefined) {
```
kevglass, line 314: Did we lose input delay configuration?

packages/dusk-cli/src/lib/validateGameFiles.ts
kevglass: This isn't an error case now is it - undefined just means they didn't specify updatesPerSecond. I think it might be worth adding a check for have they specified an update function and no updatesPerSecond or vice versa.

kevglass (approved):

packages/rune-games-cli/src/lib/validateGameFiles.ts
Gongreg: We didn't, we just don't have to worry about it when running the logic inside logic runner.

packages/dusk-cli/src/lib/validateGameFiles.ts
Gongreg: Updates per second is always going to be provided, even if dev does not set it to any value.
So theoretically this check is redundant now, but keeping it in case we change something in the future.

PR #402, opened 2024-06-18
docs: explain potential in quick start

bfelbo:
Some devs thought based on quick start that Rune only supports turn-based games so added a bit to make that more clear.

### DIFF
--- a/docs/docs/quick-start.md
+++ b/docs/docs/quick-start.md
@@ -97,7 +97,11 @@ You can find additional information about [rendering here](how-it-works/syncing-
 
 ## What Next? {#next-steps}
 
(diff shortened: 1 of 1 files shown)


Gongreg (approved):

docs/docs/quick-start.md
```diff
@@ -97,7 +97,11 @@ You can find additional information about [rendering here](how-it-works/syncing-
 
 ## What Next? {#next-steps}
 
+You can also make much fancier games than Tic Tac Toe. Rune supports games with real-time multiplayer, complex physics, beautiful graphics and much more!
+
+Here's some good places to start:
```
Gongreg: I think it should be "here are some"

PR #431, opened 2024-07-01
fix: Update interpolator to use future state rather than half way

kevglass:
Update interpolator to use future state rather than half way
Add tests for multiple axis movement


packages/interpolators/test/interpolatorLatency.spec.ts
```diff
@@ -68,17 +68,26 @@ describe("interpolator", () => {
...
     instance.update({ game: 12, futureGame: 15 })
-    expect(instance.getPosition()).toEqual(2)
+    expect(instance.getPosition()).toEqual(7)
 
+    // we're still moving in the same direction and have accelerated to 1.5 (future speed = 2)
+    // we're half way between (1.5+1.5) and (12.5 + 2) - 12.5+2 is bigger than future game, so we'll use the future
+    // game state of 14. 50% between 3 and 14 = 8.5
     instance.update({ game: 14, futureGame: 14 })
-    expect(instance.getPosition()).toEqual(3.75)
+    expect(instance.getPosition()).toEqual(8.5)
```
bfelbo: FFTI: Might be nice to finish the test by calling a `instance.update(...)` a few times and see that it accelerates to the max speed (and not above that) + gets to the actual intended position.

packages/interpolators/test/interpolatorLatency.spec.ts
```diff
@@ -68,17 +68,26 @@ describe("interpolator", () => {
...
 
+    // we're still moving in the same direction and have accelerated to 1 (future speed = 1.5)
+    // we're half way between (0.5+1) and (11 + 1.5) = 5.5 + 1.5 = 7
     instance.update({ game: 12, futureGame: 15 })
-    expect(instance.getPosition()).toEqual(2)
+    expect(instance.getPosition()).toEqual(7)
 
+    // we're still moving in the same direction and have accelerated to 1.5 (future speed = 2)
+    // we're half way between (1.5+1.5) and (12.5 + 2) - 12.5+2 is bigger than future game, so we'll use the future
+    // game state of 14. 50% between 3 and 14 = 8.5
```
bfelbo: It seems like the actual position in logic goes from 15 to 14, i.e. the player moves in the opposite direction. Is that intentional?

If so, would be great to mention that in the comment. Since the test is already quite complex, it might be better to only do the acceleration + move in opposite direction in another test.

bfelbo (approved):
LGTM 👍 Haven't gone super deep though, there might be something I haven't caught.

packages/interpolators/test/interpolatorLatency.spec.ts
bfelbo: It might be nice to just keep this test simple and do something like:
```
update({ game: 2, futureGame: 4 })
update({ game: 4, futureGame: 6 })
update({ game: 6, futureGame: 8 })
...
```

so this test focuses specifically on the interpolating using `timeSinceLastUpdate` and not anything else.

kevglass: It wasn't intentional. Fixed.

PR #445, opened 2024-07-10
feat: Update latency interpolators to maintain speed for points independently

kevglass:
Update the latency based interpolators to manage speed and acceleration independently for components/axis to allow deceleration when changing direction.


packages/interpolators/src/interpolators/interpolatorLatency.ts
```diff
@@ -51,89 +23,127 @@ export function interpolatorLatency<Dimensions extends number | number[]>({
...
+      return (
+        Math.sign(point.target - point.current) *
+        Math.min(Math.abs(maxSpeed), Math.abs(point.target - point.current)) *
+        componentCoefficient
+      )
     }
 
-    if (target > interpolated) {
+    // otherwise determine what speed the next steps should be at by applying
+    // acceleration (or deceleration). Don't even let the speed take us past
```
Gongreg:
```suggestion
    // acceleration (or deceleration). Don't ever let the speed take us past
```

PR #488, opened 2024-07-30
feat: tech demo platformer

kevglass:


tech-demos/platformer/vite.config.ts
```diff
@@ -0,0 +1,17 @@
+import path from "node:path"
+import { defineConfig } from "vite"
+import dusk from "vite-plugin-dusk"
```
Gongreg: Switch to dusk-games-sdk/vite & remove the dependency.

PR #508, opened 2024-08-02
feat: detect assets

Gongreg:
Do not allow asset files to be imported inside logic.


kevglass:
This is going to break some of my games. I use SVG for instance as an input to build game physics models for Scorched Turf. Think this would be nice to be able to turn off.

bfelbo:
AFAIK, the SVG in Scorched Turf is inlined into logic.js. Perhaps we can make the detection more intelligent to not flag inlined assets and explain in the error msg how to inline assets into logic.js if needed for that game?

bfelbo:
I.e. I'd suggest making the detection more intelligent rather than adding a flag :)

Gongreg:
@bfelbo , we've looked into Kev example and in his case there is no error/warning there. Currently asset detection is very basic, so in case user uses inline param, we won't understand that this is an asset, because `png?inline` !== `png`.

bfelbo:
Sounds great 👍

PR #520, opened 2024-08-13
feat: vite plugin check only start of string for ignored dependencies

dmecke:

### DIFF
--- a/packages/vite-plugin-dusk/src/plugins/detectExternalImports.ts
+++ b/packages/vite-plugin-dusk/src/plugins/detectExternalImports.ts
@@ -143,7 +143,7 @@ File: ${withOptionalExtension} was imported by:\n${importPath.join("\n")}
 
       const isInWhitelist =
         DEPENDENCY_WHITELIST.includes(name) ||
-        options.ignoredDependencies?.includes(name)
+        options.ignoredDependencies?.some(ignoredDependency => name.startsWith(ignoredDependency))
 
       if (skipAllowedDependencies && isInWhitelist) {
         return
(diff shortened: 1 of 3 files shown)


packages/vite-plugin-dusk/test/externalDependencies.spec.ts
```diff
@@ -113,6 +113,25 @@ describe("external dependencies", () => {
     expect(logicChunk?.code).toContain("/*! Imported dependencies: math-sum*/")
   })
 
+  it("should ignore dependencies passed in plugin options, matching only the start of the import", async () => {
+    const logger = createLogger()
+
+    const output = await buildFixture("external-basic", logger, {
+      ignoredDependencies: ["math-"],
```
dmecke: Not the best use case, but I wanted to avoid adding bloat to the fixtures or even dependencies. I can change this of course to something else if preferred.

packages/vite-plugin-dusk/src/plugins/detectExternalImports.ts
```diff
@@ -143,7 +143,7 @@ File: ${withOptionalExtension} was imported by:\n${importPath.join("\n")}
 
       const isInWhitelist =
         DEPENDENCY_WHITELIST.includes(name) ||
-        options.ignoredDependencies?.includes(name)
```
Gongreg, line 146: Hey @dmecke. Thank you for your contribution! I've assumed from the issue you've experienced, that we'd want to allow imports of multiple files from one dependency:

For instance if we allow dependency `lodash`, then importing `lodash/x`, `lodash/y` is fine, but importing a dependency `lodash-different-package` should not be allowed.

Could you update the code so that we allow package + package/inner/files ?

dmecke: I've changed the code so that it should allow packages that match the exact name as well as imports of files within this dependency.

I deleted the spec, because the one I added doesn't make sense anymore and I think there is no dependency that matches this use case that I can use for a test case.

Let me know if any further changes are needed.

Gongreg: All good, I've updated the tests. Thank you :)

Gongreg (approved):

PR #530, opened 2024-09-02
docs: Provide a simplified title as the social preview title in og:title

kevglass:
Customize the meta data provided to docusaurus to make `og:title` a short form of the actual page title.


bfelbo (approved):
👍

PR #567, opened 2024-10-11
Hide Top 10 games when necessary

helios1138:


bfelbo (approved):
💪

docs/static/img/home/topGamesHidden.png
mieszko4, line 1: This img is over 1MB. Can you compress and use webp?

Or will docusaurus do that?

mieszko4 (approved):

docs/static/img/home/topGamesHidden.png
helios1138: I can optimize the PNG, but optimization will destroy the smooth gradients
Screenshot 2024-10-11 at 20 10 47

helios1138: and webp is not 100% supported yet so I don't think we can use that

mieszko4: I think it is better for user to load some kind of img then seeing it being loaded. But FFTI

mieszko4: I've been using webp already. It is supported in all modern browsers.

helios1138: Yeah, you're right, I was confused because https://caniuse.com/webp shows latest browser version release date, not earliest supported. Changed to webp https://github.com/rune/rune/pull/567/commits/5a00c6e049628394f0500d802b8dbfabdf520e62

PR #581, opened 2024-10-29
docs: simplify play time for creator fund

bfelbo:
Write play time credits per hour instead to simplify explanation.


sanjaypojo (approved):
🙌🏽

PR #597, opened 2024-11-08
feat: Update allowed package size for games

kevglass:


bfelbo (approved):
Sounds good, let's also update the docs :)

PR #616, opened 2024-12-05
docs: Update API docs to show the request ID

kevglass:

### DIFF
--- a/docs/docs/advanced/ai.md
+++ b/docs/docs/advanced/ai.md
@@ -23,7 +23,7 @@ Example of a single plain-text prompt:
 Rune.initLogic({
   actions: {
     myAction: (payload, { game }) => {
-      Rune.ai.promptRequest({
+      game.requestId = Rune.ai.promptRequest({
         messages: [{ role: "user", content: "What is Rune.ai?" }],
       })
     },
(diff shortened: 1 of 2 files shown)


bfelbo:
Might be simpler to only have `requestId` in the API reference for request and response since it's not actually used for the other examples? :)

kevglass:
> Might be simpler to only have `requestId` in the API reference for request and response since it's not actually used for the other examples? :)

Or make the examples use it - I really think it needs to be visible in the examples

bfelbo:
Why? Doesn't seem to be needed for e.g. Storyteller AI and Dating AI Game? (didn't look at the other games)

Don't feel strongly though :)

kevglass:
Miloz first reaction was how do i track the request - absolutely, I haven't used it at all. So maybe it just shouldn't be in the API at all?

bfelbo:
IMO it's fine to have in the API so the API is ready for us adding support for multiple requests at once if needed in the future. Some devs might also prefer to store by request ID instead of pushing it to an array.

My suggestion would be to just update this PR to not mention it in `ai.md` and call it a day for now. If there's any future confusion, we can adjust it at that time

mzbyszynski:
closing this for now

PR #628, opened 2025-04-28
docs: add quickstart docs for Rune MCP in VSCode

sanjaypojo:


packages/rune-mcp/README.md
```diff
@@ -32,16 +19,39 @@ The server integrates with VS Code and GitHub Copilot to provide an enhanced dev
...
    }
    ```
-Once this is done you should see the option to refresh agents option in the GitHub Copilot chat window.
-This option only appears when you are in agent mode. 
-We recommend using the Claude 3.7 Sonnet model, but feel free to experiment with any models you would like.
+- When you save the `mcp.json` file, you should see a little "Start" button appear:
+   <p align="center">
+   <img width="796" alt="Rune mcp.json setup in VSCode" src="https://github.com/user-attachments/assets/524463d6-71b6-4b9c-ac8b-2c5c15ea8106" />
+   </p>
+- Now press "Start" and then open up the Copilot window.
```
mzbyszynski, line 26: This step isn't really necessary. If they skip it, they may see a blue refresh button on the chat window. if they press that it will start the mcp server and load the tools. The only reason I like that workflow a little better is they may think that they need to manually "start" the mcp server every time they open vs code, but really they just need to refresh the chat window one time, and then after that the agent will start the server automatically when it needs it. 

But if you think starting the server as described is still the easiest workflow, then FFTI

Screenshot 2025-04-28 at 5 28 34 PM

mzbyszynski (approved):

PR #656, opened 2026-07-06
fix: no-global-scope-mutation falsely flags rebinding a local initialized from a global

slugoguls:
`let minScore = Infinity` made the rule treat the local as an alias of the global `Infinity`, so any later `minScore = score` was flagged as a parent-scope mutation (same for `NaN`/`undefined` initializers). Rebinding never mutates what the initializer referenced, so alias resolution now only applies to object mutations (`a.x = 1`, `a.push()`, `Object.assign(a, …)`), which are still flagged.

Added 8 valid + 2 invalid regression cases:

```
✔ () => { let hest = Infinity; hest = 1 }        (+7 more, previously false positives)
✔ let hest = Infinity; (() => { hest = 1 })()    (genuine mutation, still flagged)

856 passing (920ms), 0 failing
```


slugoguls (commented):

PR #657, opened 2026-08-05
fix(vite-plugin-rune): support both rune-sdk asset layouts (root and dist/)

sazuyakun:
## What's broken

rune-sdk 6.0.6 moved `multiplayer.js` and `multiplayer-dev.js` from the package root into `dist/`. The plugin still assumes the root in two places:

- `npm run dev` crashes with `ENOENT: ... /rune-sdk/multiplayer-dev.js`
- builds inject `cdn.jsdelivr.net/npm/rune-sdk@6.0.6/multiplayer.js`, which 404s — no error, the game just never loads

## Why not just add `dist/` to the path

`peerDependencies` is `rune-sdk: >=6.0.1`, which spans both layouts, so hardcoding `dist/` fixes 6.0.6 and breaks 6.0.1.

Note that 6.0.1 ships a `dist/` as well (`logicRunner`, `gameClient`, ...) — it just keeps the browser bundles at the root. So the presence of the directory tells you nothing; you have to look for the file itself.

## Change

`resolveSdkAsset()` probes `dist/<file>` then `<file>` and returns the absolute path plus the package-relative subpath. `devPlugins` uses it to read the dev runtime, `transformHtml` to build the CDN url. Peer range left alone.

## Testing

11 new tests run both plugins against synthetic 6.0.1 and 6.0.6 packages. Reverting either plugin to the old code fails exactly the 6.0.6 cases and nothing else.

Also checked end to end, with the plugin installed from a tarball so resolution happens inside the test project:

| plugin | sdk | `GET /@rune-sdk` | build url |
| --- | --- | --- | --- |
| published 1.1.0 | 6.0.6 | 500 ENOENT | — |
| this branch | 6.0.6 | 200 | `@6.0.6/dist/multiplayer.js` |
| this branch | 6.0.1 | 200 | `@6.0.1/multiplayer.js` |

Both urls verified against jsdelivr: `@6.0.6/multiplayer.js` 404s, `@6.0.6/dist/multiplayer.js` and `@6.0.1/multiplayer.js` return 200.

This shipped through a green CI because `devOutput.spec.ts` only checks html tag order and never calls the `load()` hook that crashed. The new specs call the hooks directly.

## Also here: `.node-version` v18 → v22

Unrelated to the SDK fix, but `test-docs` fails on every PR without it, so it's folded in rather than left as a blocker.

`docs/scripts/build-examples.sh` runs a bare `yarn` in each tech-demo, and unlike the examples, none of them have a lockfile — so they resolve fresh every run. `brace-expansion@5.0.9` (published 2026-07-30) requires node 20+ and broke the build. Committing lockfiles doesn't fix it: pinning the tree just moves the error to `eslint-visitor-keys@5.0.1`, which wants `^20.19.0 || ^22.13.0 || >=24`. Node 18 went EOL in April 2025.

On node 22, `yarn build`, `yarn typecheck`, `yarn lint`, `yarn test` and all five tech-demo installs pass. v22 rather than v20 because of that `^20.19.0` constraint.

Tech-demo lockfiles are still worth adding for reproducibility, but that's a separate change.

## Notes

- This doesn't reach npm until a version bump lands — `lerna publish from-package` doesn't bump on its own.
- `latest` is currently pointed at 6.0.1, so fresh `^6.0.1` installs work today. Lockfiles already on 6.0.6 don't.
- `packageVersionCheck` compares against the `latest` tag with `!==`, so right now it tells 6.0.6 users to install 6.0.1. Pre-existing, separate issue.

## Not in this PR

forge's vendored copy of this plugin (same bug), `validateGameFiles.ts` rejecting `cdn.rune.ai` urls, anything in tango.


sazuyakun:

sanjaypojo (approved):
LGTM 🚀

PR #658, opened 2026-08-05
ci: run on node 22

sazuyakun:
## Problem

`test-docs` fails on every PR right now:

```
info No lockfile found.
error brace-expansion@5.0.9: The engine "node" is incompatible with this module.
      Expected version "20 || >=22". Got "18.20.8"
```

`docs/scripts/build-examples.sh` runs a bare `yarn` in each example and tech-demo. The seven `examples/` have committed lockfiles; the five `tech-demos/` do not, so they resolve fresh on every run and pick up whatever was published since. `brace-expansion@5.0.9` landed on 2026-07-30 and requires node 20+.

`staging` looks green only because its last CI run was 2025-11-03, before that release.

## Why not just commit lockfiles

Tried it. Pinning the tree moves the failure rather than removing it:

```
error eslint-visitor-keys@5.0.1: Expected "^20.19.0 || ^22.13.0 || >=24". Got "18.20.8"
```

More than one package in that tree has dropped node 18, which reached end of life in April 2025. Lockfiles are still worth adding for reproducibility, but they are a separate change and they do not fix this.

## Change

`.node-version` v18 → v22.

## Testing

On node 22, at the repo root: `yarn build`, `yarn rune --version`, `yarn typecheck`, `yarn lint`, `yarn test` (43 tests across 13 suites) all pass.

All five previously failing tech-demo installs succeed:

```
OK  tech-demos/top-down-synchronization
OK  tech-demos/platformer
OK  tech-demos/phaser
OK  tech-demos/threejs
OK  tech-demos/world-time
```

v22 rather than v20 because `eslint-visitor-keys@5.0.1` needs `^20.19.0` specifically, so a v20 pin would depend on setup-node resolving a recent enough patch.


sazuyakun:
Folded into #657 — it's a one-line change and #657's CI needs it to go green.

